PR #2, opened 2018-09-14
added a dev script that builds and runs a vanilla shell w/ mcfly preloaded

marshall:
This makes it simple to build changes and test them locally without injecting a dev / debug build of mcfly into your primary shell environment.

To use it, simply execute the new `dev.bash` script, and you will be dropped into a vanilla bash shell with mcfly bindings pre-wired through the `mcfly-bash.sh` init script


cantino:
Slick, thank you!

PR #5, opened 2018-12-04
Use clobbering redirects.

gwk:
Bash's `set -o noclobber` prevents `>` redirects from overwriting files. Use `>|` overwriting redirects so that mcfly works even when the user has set this option.


cantino:
Thanks @gwk!

PR #7, opened 2018-12-04
Fix tr behavior on macOS.

f355:
With `LC_CTYPE=C` it fails with `tr: Illegal byte sequence`. `LC_ALL=C` seems to work fine.


cantino:
Thanks!

PR #8, opened 2018-12-04
[WIP] Adds mcfly.zsh

grrowl:
Adds a `mcfly.zsh`. In theory it should work as a zsh widget bound to `^R`, but when I run it it throws:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 19, kind: Other, message: "Operation not supported by device" }', libcore/result.rs:1009:5
```

possibly because `~/.zsh_history` is `-rw-------` by default? it seems to have the same permissions as my `~/.bash_history`

I'm not a zsh wizard by any means, so I've probably missed things, but maybe this gives others a starting point to fix my oversights 😀

### DIFF
--- a/mcfly.zsh
+++ b/mcfly.zsh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# Avoid loading this file more than once
+if [[ "$__MCFLY_LOADED" == "loaded" ]]; then
+  return 0
+fi
+__MCFLY_LOADED="loaded"
+export MCFLY_SESSION_ID=$(cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 24 | head -n 1)
+export MCFLY_HISTORY=$(mktemp -t mcfly.XXXX)
+export HISTFILE="${HISTFILE:-$HOME/.zsh_history}"
+
+if [[ -f "$HISTFILE" ]];
+then
+  tail -n100 ${HISTFILE} >| ${MCFLY_HISTORY}
+else
+  printf "Welcome to McFly\n" >| ${MCFLY_HISTORY}
+fi
+
+# Ignore commands with a leading space
+export HISTCONTROL="ignorespace"
+
+# Append new history items to .zsh_history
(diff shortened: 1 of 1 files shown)


mcfly.zsh
```diff
@@ -0,0 +1,54 @@
...
+
+if [[ -f "$HISTFILE" ]];
+then
+  tail -n100 ${HISTFILE} >| ${MCFLY_HISTORY}
+else
+  printf "Welcome to McFly\n" >| ${MCFLY_HISTORY}
+fi
+
+# Ignore commands with a leading space
+export HISTCONTROL="ignorespace"
```
grrowl: probably bash specific

cantino:
You also may need to add a new method of parsing the history file in `mcfly` if zsh's history format isn't the same as bash's (1 command per line, optionally with a commented out timestamp on the line before).

mcfly.zsh
scalp42: Equivalent would be:

```
setopt HIST_IGNORE_SPACE
```

scalp42: Equivalent would be:

```
setopt HIST_IGNORE_SPACE
```

cole-h:
I am a beginner in Rust, but thought I'd take a crack at the Rust parsing of the `zsh` history file. It is rather rudimentary at the moment, but I'll clean it up and send in a pull tomorrow afternoon-ish.

If I don't get around to it and somebody else decides to implement this: don't forget that `zsh` has the `EXTENDED_HISTORY` option which puts the current Unix epoch and runtime of the command _on the same line_ as the command. Example:

`: 1458291931:0;ls -l`

mcfly.zsh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
ericbn: You don't want to have a bash shebang for a zsh script!

grrowl:
Sorry I'm going to close this, but anyone is free (with more zsh scripting experience perhaps?) to pick up where I left off :)

PR #26, opened 2018-12-07
Support spaces in HISTFILE

markusjevringgoeuro:
There was one remaining place that wasn't quoted in the script


cantino:
Thanks @markusjevringgoeuro!

PR #33, opened 2018-12-26
fix 'cat /dev/urandom' not closing

### LINKED ISSUE #32: cat /dev/urandom never closes
I'm trying to build a package for Nixos, but when I source the macfly.bash file, it hangs on the following line:

https://github.com/cantino/mcfly/blob/a548e310152ec036a06625ba5521d51ad0ddd25c/mcfly.bash#L8

It's maybe due to the problem described [in this forum](https://www.unix.com/unix-for-advanced-and-expert-users/271649-cmd-cat-dev-urandom-not-closing-cleanly.html). Cat is never stopped...

What seems to stop the command is using `head -c 24` instead of the `fold ... | head -n 1`:

```
[sam@Waluigi:~/projects/mcfly]$ x=$(cat /dev/urandom | env LC_ALL=C tr -dc 'a-zA-Z0-9' | head -c 24)
tr: write error: Broken pipe
tr: write error
cat: write error: Broken pipe
[sam@Waluigi:~/projects/mcfly]$ echo $x
ktgQD1V2qVWbvuLUnFTEhwGg
```

But then the broken pipe errors appear as you can see. I think they could be discarded with
```
x=$(cat /dev/urandom 2>/dev/null| env LC_ALL=C tr -dc 'a-zA-Z0-9' 2>/dev/null | head -c 24)
```

Melkor333:
Fix #32 by letting head close the pipe after 24 characters which seems more stable

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -5,7 +5,7 @@ if [[ "$__MCFLY_LOADED" == "loaded" ]]; then
   return 0
 fi
 __MCFLY_LOADED="loaded"
-export MCFLY_SESSION_ID=$(cat /dev/urandom | env LC_ALL=C tr -dc 'a-zA-Z0-9' | fold -w 24 | head -n 1)
+export MCFLY_SESSION_ID=$(cat /dev/urandom | env LC_ALL=C tr -dc 'a-zA-Z0-9' | head -c 24)
 export MCFLY_HISTORY=$(mktemp -t mcfly.XXXX)
 export HISTFILE="${HISTFILE:-$HOME/.bash_history}"
 


cantino:
Works for me on Mac and Linux. Thanks @Melkor333!

karsibali:
This app does not get along with vs code and hyper either. Take a look at the following issues:
[https://github.com/zeit/hyper/issues/4465](url)
[https://github.com/microsoft/vscode-cpptools/issues/5275](url)

cantino:
Looks like the general fixes worked for you: https://github.com/cantino/mcfly/issues/32#issuecomment-619292658

PR #34, opened 2019-01-18
make file check more precise

camerondavison:
-r is TRUE when file is readable


camerondavison:
not a big deal, just thought it could be nice.

cantino:
Thanks :)

PR #35, opened 2019-01-23
Use #is_digit to check if char is a digit

teeparham:
https://doc.rust-lang.org/std/primitive.char.html#method.is_digit


cantino:
Thanks!

PR #52, opened 2019-03-24
Fix clippy errors

winding-lines:
Adjusted the code as per clippy's recommendation. In some case I
disabled the clippy check.


cantino:
Thanks!

src/history/history.rs
```diff
@@ -168,48 +169,50 @@ impl History {
...
-                                } else {
-                                    // The source must have been a file, so ignore it.
-                                }
-                                return;
+    pub fn possibly_update_paths(&self, command: &str, exit_code: Option<i32>) {
+        let successful = exit_code.is_none() || exit_code.unwrap() == 0;
+        let is_move = |c:&str| c.to_lowercase().starts_with("mv ")
+            && !c.contains('*')
+            && !c.contains('?');
+        if  successful && is_move(command) {
```
cantino: Was this because clippy complained about nested `if`s?

Also, extra whitespace `if  successful`

src/bash_history.rs
```diff
@@ -7,12 +7,13 @@ use std::fs::File;
 use std::io::Read;
 
 fn read_ignoring_utf_errors(path: &PathBuf) -> String {
-    let mut f = File::open(path).expect(format!("McFly error: {:?} file not found", &path).as_str());
+    let mut f = File::open(path).unwrap_or_else(|_| panic!("McFly error: {:?} file not found", &path));
```
cantino: All the switched to `unwrap_or_else` are good-- they'll be more performant, thanks.

src/history/history.rs
winding-lines: Correct, clippy doesn't like nested `if`.

winding-lines: I did git amend and force push to fix the white space :)

I am not sure how you feel about `rustfmt`, I use it on some of my projects. There is a one time change that affects all the files.

cantino: I don't have a strong opinion, being fairly new to rust. Would you recommend it?

winding-lines: I like not to have to worry about formatting, specially if we start to get contributions :)

cantino: Ok, let's do it! 😄(In a new PR, please).

cantino:
Thanks!

PR #54, opened 2019-04-02
Only read 256 bytes for session id generation

SuperSandro2000:
I tried [ttyd](https://github.com/tsl0922/ttyd) today and ran into these error messages.

I traced it down to mcfly's session id generation. 
As far as I understood the pipe overflows in this scenario and then throws that error despite still generating the id.
It still annoyed me so I searched for a fix and found one by limiting the input bytes to 256.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -17,7 +17,7 @@ if [[ ! -r "${HISTFILE}" ]]; then
 fi
 
 # MCFLY_SESSION_ID is used by McFly internally to keep track of the commands from a particular terminal session.
-export MCFLY_SESSION_ID=$(cat /dev/urandom | env LC_ALL=C tr -dc 'a-zA-Z0-9' | head -c 24)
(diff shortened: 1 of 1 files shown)


cantino:
Thank you for the fix!

nfultz:
I hit the same issue, but a slightly different fix - https://github.com/nfultz/mcfly/pull/1/commits/b3189ccc4d61910315406bc76c7f1da79fc1ed79

I don't understand why I only got the error on JupyterLab's terminal, and not a regular one.

SuperSandro2000:
The regular terminal cuts off the random stream after a while.
Maybe with pipefail set it will fail, too.

cantino:
Sorry for the delay here. Does this mean that we need a further fix @nfultz @SuperSandro2000?

SuperSandro2000:
Don't think so. I was just answering why maybe some terminals fail and others don't

nfultz:
I think my way is somewhat cleaner, but not sure if the `base64` program is portable (I don't test on mac or windows). No change req'd afaict.

PR #57, opened 2019-04-10
F4 toggles folder

winding-lines:
Hitting F4 will toggle the folder as part of the command

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -22,6 +22,7 @@ pub struct Interface<'a> {
     debug: bool,
     run: bool,
     menu_mode: MenuMode,
+    display_mode: DisplayMode,
 }
 
 pub struct SelectionResult {
@@ -43,7 +44,7 @@ pub enum MenuMode {
 impl MenuMode {
     fn text(&self) -> &str {
         match *self {
-            MenuMode::Normal => "McFly | ESC - Exit | ⏎ - Run | TAB - Edit | F2 - Delete",
+            MenuMode::Normal => "McFly | ESC - Exit | ⏎ - Run | TAB - Edit | F2 - Delete | F4 - Toggle folder",
             MenuMode::ConfirmDelete => "Delete selected command from the history? (Y/N)",
         }
     }
@@ -54,6 +55,12 @@ impl MenuMode {
             MenuMode::ConfirmDelete => color::Bg(color::Red).to_string(),
         }
     }
+}
+
+#[derive(PartialEq)]
+pub enum DisplayMode {
+    Command,
+    WithPath,
 }
 
 const PROMPT_LINE_INDEX: u16 = 3;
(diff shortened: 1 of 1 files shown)


aik099:
What exactly does "toggle the folder" action mean. Could you please show some examples of what happens when the F4 is pressed (compared to current behavior) for various commands from history (e.g. when past command selected, when past command not selected)?

winding-lines:
Alex,

Thanks for checking this out :) This is my first actual contribution to your project and I am looking for feedback at all levels:

- how to name menu entries
- behavior
- implementation

In the current implementation when you hit F4 all the displayed commands get prepended with "cd <folder> && <actual command>". Hitting F4 again will remove the cd. The current filter text does not change.

Screen Shot 2019-04-10 at 9 44 26 AM
Screen Shot 2019-04-10 at 9 44 35 AM

Note that in the list there is an earlier command that I used with F4 so it already has a cd. The behavior in that case is not intuitive: a new cd gets appended. One possible approach is to
 not prepend "cd <folder> &&" if the command already starts with "cd /" or "cd ~/".

Let me know which way I should take this :)

aik099:
What is the use case of proposed functionality? The McFly already suggests it’s result based on a current directory.

winding-lines:
The use case is for when you are switching projects. I frequently switch projects in the same day. For example I may run ‘code .’ from variois folders. Or run ‘git pull -rebase’. By changing the folder to the one of the other project I change context with one command.

Let me know what you think :-)
---

aik099:
So you're in `project_two` folder and seeing command history from it and then press F4 and command from `project_two`, that you've chosen is applied to `project_one`.

Seems very specific use case. But we'll wait for @cantino to review.

winding-lines:
Sorry, I was not able to explain. Let me be more verbose and setup more context.

I have multiple projects that I work on. For example I may be working on a machine learning project and then I want to change mcfly for a bit, then I need to adjust my dotfiles. I generally use vim for editing the smaller project or (vs)code for the bigger ones. I prefer to focus on one project at a time.

So I am in my squad project and I am using mcfly to repeatedly run commands in this one project: Ctrl-R for mcfly, search, select then repeat.

Now I want to work in mcfly. I know that I generally do ‘git fetch upstream’ when i switch to mcfly. So I do Ctrl-R for mcfly, search git, and I find the command I need. Now I hit F4 and the command in mcfly changes to ‘cd ~/dev/mcfly && git fetch upstream’. I select that command and I changed context while at the same time bootstraping my workflow in the new project.

Does this make more sense? Many thanks for asking more clarifying questions if it doesn’t. 

It’s also ok if mcfly does not want this feature at all. It’s something that I use multiple times a day and I am happy to keep it in my fork. We could maybe split a shared library in that case.



Marius
---

cantino:
Hey @winding-lines, thank you for the contribution! I understand your use-case, but wonder if it really makes sense to be part of McFly. For the same use-case, I generally do `j mcfly` (using [autojump](https://github.com/wting/autojump)) and then run whatever commands using McFly. I don't know that having this built into McFly will be useful for enough people, but I'm open to opposing views if other users would like to chime in and let us know their opinions!

winding-lines:
Thanks for checking it out :-)

Marius
---

PR #73, opened 2019-11-29
Implement Vim key scheme

### LINKED ISSUE #72: Alternate Vim keybindings
As a fairly frequent Vim user I often find that I open mcfly and immediately try to use `h`/`j`/`k`/`l` to move the cursor up and down. Therefore a nice feature might be alternate key bindings that more closely matches vim:
- `j`/`k` - move up/down
- `i` - enter/exit edit mode
- `ESC` - exit edit mode (and exit mcfly?)
- `q` - exit mcfly?

This would be set with an environment variable:
```
MCFLY_KEYBINDINGS=vim
```

JamJar00:
Implements a base set of Vim keybindings in McFly, enabled optionally by setting the `MCFLY_KEY_SCHEME` environment variable to `vim`. Closes #72.

Cheatsheet for Vim keys [here](https://devhints.io/vim) if you're not familiar. Only a subset has been implemented, some have been subtly simplified (i.e. `:q` is just `q`) and other functionality ommited entirely for the sake of simplicity (e.g. `3x`).

I'm not a Rust programmer sadly* so I appreciate any and all feedback, suggestions, code quality fixes, etc. you can provide. I've tried to do the simplest solutions I could think of for the most part so I'm more than happy to rearchitect parts in the 'Rust' way if there's better choices that could have been made!

\* This was mostly built by flailing around trying to get easier errors out of the compiler

Jamie

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -408,16 +408,6 @@ impl<'a> Interface<'a> {
                     self.menu_mode = MenuMode::ConfirmDelete;
                 }
             }
-            Key::Ctrl(_c) => {
-                //                      self.debug(&mut screen, format!("Ctrl({})", c))
-            }
-            Key::Alt(_c) => {
-                //                      self.debug(&mut screen, format!("Alt({})", c))
-            }
-            Key::F(_c) => {
-                //                      self.debug(&mut screen, format!("F({})", c))
-            }
-            Key::Insert | Key::Null | Key::__IsNotComplete => {}
             _ => {}
         }
 
@@ -469,16 +459,6 @@ impl<'a> Interface<'a> {
                         self.menu_mode = MenuMode::ConfirmDelete;
                     }
                 }
-                Key::Ctrl(_c) => {
-                    //                      self.debug(&mut screen, format!("Ctrl({})", c))
-                }
-                Key::Alt(_c) => {
-                    //                      self.debug(&mut screen, format!("Alt({})", c))
-                }
-                Key::F(_c) => {
-                    //                      self.debug(&mut screen, format!("F({})", c))
-                }
-                Key::Insert | Key::Null | Key::__IsNotComplete => {}
                 _ => {}
             }
         } else {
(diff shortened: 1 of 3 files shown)


cantino:
Awesome @JamJar00!

src/interface.rs
```diff
@@ -400,6 +338,216 @@ impl<'a> Interface<'a> {
...
+                //                      self.debug(&mut screen, format!("Ctrl({})", c))
+            }
+            Key::Alt(_c) => {
+                //                      self.debug(&mut screen, format!("Alt({})", c))
+            }
+            Key::F(_c) => {
+                //                      self.debug(&mut screen, format!("F({})", c))
+            }
+            Key::Insert | Key::Null | Key::__IsNotComplete => {}
+            _ => {}
```
cantino: Does it error if you don't have this line? I thought `Key::Char(c)`, `Key::Ctrl(_c)`, `Key::Alt(_c)`, `Key::F(_c)` covered the match.

src/interface.rs
```diff
@@ -41,9 +43,16 @@ pub enum MenuMode {
 }
 
 impl MenuMode {
-    fn text(&self) -> &str {
+    fn text(&self, interface: &Interface) -> &str {
         match *self {
-            MenuMode::Normal => "McFly | ESC - Exit | ⏎ - Run | TAB - Edit | F2 - Delete",
+            MenuMode::Normal => match interface.settings.key_scheme {
```
cantino: Can this just be `self.settings.key_scheme` so that you don't need to pass `interface` in?

src/interface.rs
```diff
@@ -400,6 +338,216 @@ impl<'a> Interface<'a> {
...
+                self.input.insert(c);
+                self.refresh_matches();
+            }
+            Key::F(2) => {
+                if !self.matches.is_empty() {
+                    self.menu_mode = MenuMode::ConfirmDelete;
+                }
+            }
+            Key::Ctrl(_c) => {
+                //                      self.debug(&mut screen, format!("Ctrl({})", c))
```
cantino: You could delete these comments that I left here and everywhere else if you want.

src/interface.rs
JamJar00: It seems not. I'm not 100% but I think `self` here is of type `MenuMode`, which has no reference to `settings` because that's on `Interface`.

src/interface.rs
JamJar00: I've collapsed these all down to just `_ => {}`

JamJar00:
I've added a little bit to the readme about enabling it. I always forget to update readmes

Would you like a key map written out? It feels to me that if a user is enabling vim keybindings then they probably already know the basics enough to get by, and information is plentiful online anyway!

src/interface.rs
cantino: Yea, that makes sense.

cantino:
Looks great. I'll get this merged and pushed shortly, thank you for contributing it!

cantino:
Released in v0.3.6. Thanks!

PR #81, opened 2020-05-10
Zsh

### LINKED ISSUE #3: Zsh
Awesome!! Any plan on making it available for zsh?

### LINKED ISSUE #64: Do not overwrite HISTFILE
I am using a custom HISTFILE name. And to keep it that way, I also make this variable readonly.

As mcfly does
`export HISTFILE="${HISTFILE:-$HOME/.bash_history}"`

this leads to an annoying warning:
`bash: HISTFILE: readonly variable`
A check like
```
if [ -z "${HISTFILE}" ]; then
    export HISTFILE="${HOME}/.bash_history"
fi
```
would prevent this warning for me.

cantino:
Initial work on Zsh support. I'm currently testing it locally.

Fixes #64. Fixes #3.


Flat:
From my brief usage it seems to function fine. There does seem to be some unneeded command history after ctrl+r to use mcfly and after selecting a command such as:
```
~ on  mcfly
❯ #mcfly:

~ on  mcfly
❯  mcfly search
```
It would be better if these were hidden from the terminal. Not sure if this is intentional.

cantino:
Interesting, I wasn’t getting those. What ZSH history settings are you
using?

Flat:
You can find my full .zshrc [here](https://github.com/Flat/dotfiles/blob/master/.zshrc) (with fzf replaced with mcfly of course)

Relevant? Zsh history ops:

```
setopt extended_history
setopt hist_expire_dups_first
setopt hist_ignore_dups
setopt hist_ignore_space setopt hist_verify
setopt inc_append_history
setopt share_history 
```

Haprog:
@cantino Thank you! This is awesome. I just installed this and briefly tested (I'm running `zsh` with `oh-my-zsh`). I tried few searches and ran commands. Seems to work nicely. Didn't see any issues so far.

I didn't get any extra lines in history files (those that @Flat mentioned).

Ps. I didn't even remember what is mcfly since I never tried it before. I had just subscribed to notifications on the issue about zsh support and remembered that it was something I wanted to have but couldn't before. This is a very nice improvement to my workflow as I often use Ctrl+R search and the default one had pretty bad UX.

mcfly.zsh
```diff
@@ -0,0 +1,103 @@
...
+# MCFLY_SESSION_ID is used by McFly internally to keep track of the commands from a particular terminal session.
+export MCFLY_SESSION_ID=$(dd if=/dev/urandom bs=256 count=1 2> /dev/null | env LC_ALL=C tr -dc 'a-zA-Z0-9' | head -c 24)
+
+setopt extended_history       # record timestamp of command in HISTFILE
+setopt hist_expire_dups_first # delete duplicates first when HISTFILE size exceeds HISTSIZE
+setopt hist_ignore_dups       # ignore duplicated commands history list
+setopt hist_ignore_space      # ignore commands that start with space
+setopt inc_append_history     # add commands to HISTFILE in order of execution
+setopt share_history          # share command history data
+setopt interactive_comments   # allow comments in interactive shells (like Bash does)
```
infokiller, line 28: Are all these options required for using mcfly?
If they are, I'd suggest clearly documenting that to avoid surprising users when you change their options.
If they're not, better leave the options the users set themselves.

mcfly.zsh
```diff
@@ -0,0 +1,103 @@
...
+function erase_history { local HISTSIZE=0; }
+
+# Setup a function to be used by $PROMPT_COMMAND.
+function mcfly_prompt_command {
+  local exit_code=$? # Record exit status of previous command.
+
+  # Populate McFly's temporary, per-session history file from recent commands in the shell's primary HISTFILE.
+  if [[ ! -f "${MCFLY_HISTORY}" ]]; then
+    export MCFLY_HISTORY=$(mktemp -t mcfly.XXXXXXXX)
+    tail -n100 "${HISTFILE}" >| ${MCFLY_HISTORY}
```
infokiller, line 40: It's best to use `fc` than to directly read the user history for the following reasons:
- Some commands may span multiple lines
- `fc` may contain more entries than `HISTFILE`. For example, the user may populate `fc` from multiple history files, each used for a specific machine, and sync them between machines.
So a user may have `.zsh_history_machine1` that holds the commands history typed on machine1 and `.zsh_history_machine2` that holds the commands history typed on machine2. Then he can sync them, and load them on zsh startup so that `fc` contains on each of the machines.
- You will avoid the call to `tail` which is good for performance (though probably negligible)

mcfly.zsh
```diff
@@ -0,0 +1,103 @@
...
+  # Clear the in-memory history and reload it from $MCFLY_HISTORY
+  # (to remove instances of '#mcfly: ' from the local session history).
+  erase_history
+  fc -R "${MCFLY_HISTORY}"
+  return ${exit_code} # Restore the original exit code by returning it.
+}
+
+# Set $PROMPT_COMMAND run mcfly_prompt_command and then any existing $PROMPT_COMMAND.
+PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
+precmd_functions+=(mcfly_prompt_command)
```
infokiller, line 66: I think `PROMPT_COMMAND` is bash only?
BTW, you can run a command on every prompt in zsh robustly like this:
```zsh
autoload -Uz add-zsh-hook
add-zsh-hook preexec mcfly_prompt_command
```

mcfly.zsh
```diff
@@ -0,0 +1,103 @@
...
+#   2. Type "mcfly search" and then run the command. McFly will pull the last line from the $MCFLY_HISTORY file,
+#      which should be the commented-out search from step #1. It will then remove that line from the history file and
+#      render the search UI pre-filled with it.
+if [[ $- =~ .*i.* ]]; then
+  if set -o | grep "vi " | grep -q on; then
+    bindkey -s '\C-r' '\e0i#mcfly: \e\C-j mcfly search\C-j'
+  else
+    bindkey -s '\C-r' '\C-a#mcfly: \C-j mcfly search\C-j'
+  fi
+fi
```
infokiller, line 103: You can bind keybindings more robustly in zsh than bash - no need for translating keys or reading a file. Something like this should work, assuming you can call `mcfly search --query`:

```zsh
mcfly-history-widget() {
  local selected
  selected="$(mcfly search --query="${LBUFFER}")"
  local ret=$?
  if [[ -n $selected ]]; then
    LBUFFER="$selected"
  fi
  zle reset-prompt
  return $ret
}
zle -N mcfly-history-widget
bindkey '^R' fzf-history-widget
```

cantino: Thanks, I'll try this out!

Haprog:
I'm not sure what happened. I'm very sure I didn't get those extra lines when I tried this first time but noticed now (after reboot etc.) that now I get those extra lines showing in my terminal after using Ctrl+R:

```
#mcfly:
 mcfly search
```

cantino:
Are they in your command history, or just in the terminal buffer?

Flat:
For me they seem to just be in the terminal buffer. Not in .zsh_history.

cantino:
That's the same behavior as bash. I haven't found a good way to prevent them, although @infokiller's [suggestion](https://github.com/cantino/mcfly/pull/81#discussion_r423022656) might work for zsh.

mcfly.zsh
cantino: I've now done some testing— I think the only one I actually need is `interactive_comments`.

mcfly.zsh
cantino: @infokiller, here's a question for you.

`mcfly search` can be used in two ways:

1. Just running `mcfly search` brings up the UI and looks at the last line of `MCFLY_HISTORY`. If it matches "#mcfly: foo" then it prefills the UI with "foo". The rationale for this was that, at least in bash, I couldn't figure out how to safely pass the current command to mcfly while escaping all possible special characters.
2. `mcfly search "foo"` searches for "foo"

Do you know how I could use zsh's hooks to grab the current command buffer and pass it to mcfly in a fully-escaped way without using technique 1, which is what is done right now?

Thanks for your very helpful code review!

infokiller: Perhaps I don't understand something, but in the code example above, `mcfly search "${LBUFFER}"` should pass `"${LBUFFER}"` (which is what the user typed and is to the left of the cursor) with the exact string they typed. This uses standard shell escaping.
If however you want the argument passed to mcfly to be escaped when mcfly gets it, you can use `"${(q)var}"` (which quotes with backslashes), `"${(qq)var}"` (single quotes), `"${(qqq)var}"` (double quotes). Example output:

```zsh
❯ var='a $ " \\'
echo "${(q)var}"
echo "${(qq)var}"
echo "${(qqq)var}"
a\ \$\ \"\ \\
'a $ " \'
"a \$ \" \\"
```

cantino: I think you're correct. I ran into a new problem, though. When I use your example, I get this error: `Os { code: 25, kind: Other, message: "Inappropriate ioctl for device" }`
It's coming out of the `selected="$(mcfly search --query="${LBUFFER}")"` call, and `stdout().into_raw_mode().unwrap()` in Rust. My guess is that I can't show a fullscreen UI when in this context.

cantino:
If anyone wants to test it again, I've worked with @infokiller's feedback and think I've removed the  `#mcfly: ... mcfly search` nonsense in the command history.

Haprog:
@cantino I just updated to the latest version in this branch (commit 8eecd8d5) and I don't get the `#mcfly: ... mcfly search` in terminal anymore.

Otherwise this seems to work, but now arrow keys don't work in mcfly anymore to navigate up or down the suggestions list. Esc, Tab and Enter at least work as they should (can edit or run the first suggestion easily).

Actually while writing this I just noticed that PageUp and PageDown keys work to navigate the suggestions up/down, but I would much prefer to use arrow keys for it instead. I'm pretty sure arrow keys worked for this when I was using the initial version of this (commit 5a5784a4)

cantino:
Thanks @Haprog, yes, that was definitely working before. Darn, I wonder if I can make zsh not consume those.

cantino:
Hey @Haprog, are you trying the current version? Arrow keys are working for me in zsh. Vim or emacs mode?

Haprog:
@cantino I'm running this latest `zsh` branch (as I mentioned above linking the exact commit hash). I installed it with the instructions in the readme section [Install manually from source](https://github.com/cantino/mcfly/blob/5a5784a49abf3e6b47466b553a2e4b21efe65b01/README.md#install-manually-from-source)

After having first installed the older version of this branch I just pulled latest updates (to get my branch up to date with latest changes) and ran `cargo install --path .` since I already have the relevant stuff in my PATH and in `~/.zshrc`. Is there something else I should do?

The arrow keys still don't seem to work (I've also rebooted my laptop since last try).

> By default Mcfly uses an `emacs` inspired key scheme. If you would like to switch to the `vim` inspired key scheme, set the environment variable `MCFLY_KEY_SCHEME`.

This is a bit confusing to me since I have no idea what is the difference and so I have just used the defaults without setting `MCFLY_KEY_SCHEME` manually. It would be great if there was at least some example of what keys are different when using McFly. I'm familiar with `vim` basics and do use `vim` a little bit but I've never used `emacs`.

I tried setting it to `vim` mode now but didn't notice a difference (except it says "McFly (Vim)" at the top when used), arrow keys still don't work.

Haprog:
Can I switch between different versions just by switching the binary at `~/.cargo/bin/mcfly` or does it depend on some other files?

I'm just thinking I could try compiling different versions from different commits and backup the binary after compilation to be able to switch back to better working version faster without having to run cargo install every time.

Haprog:
I just noticed that if I manually run `mcfly search` then arrow keys work as they should, but if I start it with Ctrl+R the arrow keys don't work.

Haprog:
I tested different versions of the `mcfly.zsh` file by checking out specific commits and starting a fresh terminal.

Arrow keys still work after Ctrl+R when using this commit:

- 0550b068 `Integrate comments from @infokiller` ([mcfly.zsh](https://github.com/cantino/mcfly/blob/0550b068b2cbf92666735a1644b6cc14fc813faf/mcfly.zsh))

But after the next commit the arrow keys don't work with Ctrl+R anymore:

- c3838fa1 `zsh keybindings leave clean shell` ([mcfly.zsh](https://github.com/cantino/mcfly/blob/c3838fa1097668e32e436810628f3a401558ad88/mcfly.zsh))

cantino:
Thanks @Haprog, oddly, they work for me in iTerm and Terminal on OS X. What OS / emulator are you using?

cantino:
> Can I switch between different versions just by switching the binary at `~/.cargo/bin/mcfly` or does it depend on some other files?

These changes also affected the mcfly.zsh file, so you'd need to both recompile mcfly and update to the correct version of mcfly.zsh.

Haprog:
I'm on Pop!_OS 20.04 (Ubuntu based Linux distro) using GNOME Terminal 3.36.1.1 (the default), but I also tested [Hyper](https://hyper.is/) 3.0.2 (same issue).

Haprog:
I can later test also on WSL (Windows subsystem for Linux) on Windows 10.

Haprog:
I just tested this branch also on Windows 10 running it with zsh in WSL1 (Ubuntu 20.04). Tested with two terminal apps: Command Prompt (cmd.exe, default) and the newer separately installed Windows Terminal. Both have the same issue for me that arrow keys don't work in McFly when starting it with Ctrl+R but works when started manually with command `mcfly search`.

cantino:
Thanks @Haprog for all of your testing! I'll see if I can reproduce this and figure it out.

jrolfs:
Tried **mcfly** @ 8eecd8d5fe9eb991b6aaa89c2bc5f4d3aed53fe0 with...
 
  | name | version
----:|------|--------
System | macOS | 10.15.4
Shell | zsh | 5.7.1 (x86_64-apple-darwin17.7.0)
Terminal | kitty | 0.17.1
Terminal | Terminal.app | 2.10 (433)

...and arrow keys don't work, but **`ctrl+n`** and **`ctrl+p`** do. Is that intentional? I personally prefer those bindings over the arrow keys so hopefully fixing the arrow keys doesn't make those go away ;)

Ah, I just tried invoking `mcfly search .` manually like @Haprog mentioned and I'm seeing the same behavior (the arrow keys work).

I also tried `export MCFLY_KEY_SCHEME=vim` and **`j/k`** work for up and down. This is a _little_ confusing though as I do use **`ctrl+p/n`** all the time in vim.

-----

Side note: in case anyone uses **zplug**, this is how I added mcfly:

```sh
# As a "plugin" to source key binding
zplug "cantino/mcfly", at:zsh, use:mcfly.zsh
# As a "command" to compile binary and add it to $PATH
zplug "cantino/mcfly", \
  at:zsh, \
  as:command, \
  use:"target/release/mcfly", \
  hook-build:"cargo install --path ."
```

cantino:
Thanks @jrolfs! This and the feedback from @Haprog is very helpful. I'll see if I can figure out how to fix it.

Haprog:
@cantino I guess you could try to reproduce it in a fresh Ubuntu virtual machine if you can't reproduce it on your mac directly. You can get ready to go virtual machine images of Ubuntu for VirtualBox (free) or VMware from here https://www.osboxes.org/ubuntu/

cantino:
Okay, so this has something to do with normal mode vs application mode in the terminal. https://invisible-island.net/ncurses/ncurses.faq.html#cursor_appmode

I don't really understand it. If I do `export TERM=vt220`, arrows start working for mcfly in zsh.

What's strange is that they work find even without the TERM overloaded if you just run `mcfly search` directly, so zsh is either consuming the keys, or it's changing something about the terminal mode, when running the key binding. Any terminal experts here?

cantino:
@jrolfs & @Haprog, mind testing this again?

infokiller:
@cantino I just took a look at all the comments since my last review.
Indeed, you shouldn't set `TERM`, as that will cause troubles. I would also not use `tput init`. Instead, you can try defining the following functions:

```zsh
enable-term-application-mode() { echoti smkx }
disable-term-application-mode() { echoti rmkx }
```

Then, call `disable-term-application-mode` before you launch mcfly (which may fix the arrow keys issue), and `enable-term-application-mode` afterwards.

Haprog:
> @jrolfs & @Haprog, mind testing this again?

I just quickly tested the latest version and it seems to work perfectly. Thanks for the updates!

> I have noticed that mcfly is slower to come up in zsh and then bash, and I'm not sure why.

At least I don't notice any delay when opening mcfly in zsh (though I'm on pretty high end hardware). It opens immediately for me. Maybe the slowness is only noticeable in a VM?

cantino:
> @cantino I just took a look at all the comments since my last review.
> Indeed, you shouldn't set `TERM`, as that will cause troubles. I would also not use `tput init`. Instead, you can try defining the following functions:
> 
> ```shell
> enable-term-application-mode() { echoti smkx }
> disable-term-application-mode() { echoti rmkx }
> ```
> 
> Then, call `disable-term-application-mode` before you launch mcfly (which may fix the arrow keys issue), and `enable-term-application-mode` afterwards.

Awesome, that seems to be working! Mind testing again?

cantino:
Does the new version work for you @jrolfs?

Haprog:
Still works great after the `echoti` update.

cantino:
I think this is mergeable now?

orenyomtov:
Would love to see this merged

cantino:
Done! Releasing 0.4.0 now 🥳

jrolfs:
Sorry, I missed all these notifications somehow but glad y’all figured it out. Will switch over to master when I’m back on a 💻!

Stormheg:
Late to the party but just wanted to say thanks to everyone involved for adding zsh support!

Been using mcfly for a while now and love it :heart: Thank you for creating such an amazing piece of software!

cantino:
Thanks @Stormheg ❤️

Issue #80, opened 2020-05-10
More results


dim-geo:
Hi,

is it possible to get more results? (>10)
Is this user configurable?

cantino:
Hey @dim-geo, welcome! It's currently [hardcoded here](https://github.com/cantino/mcfly/blob/master/src/interface.rs#L71) but would be easy to wire through the settings.

grahamperrin:
I tried to get 25 but I still get 10. 

Below, please, what am I doing wrong?

```text
mowa219-gjp4-8570p% tail -n 6 ~/.zshrc 
export VISUAL=/usr/local/bin/nano

if [[ -r "/usr/local/share/zsh/site-functions/_mcfly" ]]; then
  source "/usr/local/share/zsh/site-functions/_mcfly"
fi
MCFLY_RESULTS=25
mowa219-gjp4-8570p% uname -sr
FreeBSD 13.0-CURRENT
mowa219-gjp4-8570p% pkg query pkg query '%o %v %R' mcfly
Invalid query: query should contain a format string
mowa219-gjp4-8570p% pkg query '%o %v %R' mcfly 
sysutils/mcfly 0.5.2 FreeBSD
mowa219-gjp4-8570p% 
```

cantino:
Try `export MCFLY_RESULTS=25`

lengyijun:
I feel the default number should be set 25.
And the environment should be written into Readme.md

cantino:
Or maybe 20 since sometimes people use small terminal windows. If you want to send a PR to update the default, that'd be appreciated.

kentakom1213:
Since the implementation of result scrolling in #449, it might be a good idea to increase the default number of search results displayed.

cantino:
Sure, what do you think it should be?

kentakom1213:
It would be ideal if all results were visible when the terminal is full-sized, so perhaps 30 would be a good number?

cantino:
Fixed in https://github.com/cantino/mcfly/pull/453

cantino:
Released in https://github.com/cantino/mcfly/releases/tag/v0.9.3

PR #82, opened 2020-05-17
Add setting for number of results returned

freamon:
As requested in issue #80
Moved number of results from hard-coded value to something user can change via 
`mcfly search --results` option, either directly or via `mcfly.bash`

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -60,8 +60,8 @@ PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 #      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
   if set -o | grep "vi " | grep -q on; then
-    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search\C-j'"
+    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search --results 10\C-j'"
   else
-    bind "'\C-r': '\C-amcfly: \e# mcfly search\C-j'"
+    bind "'\C-r': '\C-amcfly: \e# mcfly search --results 10\C-j'"
   fi
 fi
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -60,8 +60,8 @@ PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 #      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
   if set -o | grep "vi " | grep -q on; then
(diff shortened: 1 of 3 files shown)


mcfly.bash
```diff
@@ -60,8 +60,8 @@ PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 #      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
   if set -o | grep "vi " | grep -q on; then
-    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search\C-j'"
+    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search --results 10\C-j'"
```
cantino, line 63: You can probably skip adding `--results 10` here since you have 10 as the default.

src/settings.rs
```diff
@@ -229,6 +237,9 @@ impl Settings {
                 } else {
                     settings.dir = env::var("PWD").unwrap_or_else(|err| panic!(format!("McFly error: Unable to determine current directory ({})", err)));
                 }
+                if let Ok(results) = value_t!(search_matches.value_of("results"), u16) {
```
cantino, line 240: If you want, you could fallback to a `MCFLY_RESULTS` env variable so that users can change the default easily.

cantino:
Awesome work @freamon, thank you!

mcfly.bash
freamon: True. My thinking when I amended that file was just to make it easier for the end-user using Ctrl-R (it would be clearer to know which part of the file to edit)

src/settings.rs
freamon: It seemed like a command-line switch fitted in more with existing code, as `results` only affects one subcommand, and the other MCFLY_ environment variables affect the program as a whole.

cantino: I'd suggest we add both. I'm happy to do it.

mcfly.bash
cantino: I think most people will be using the mcfly.bash file out of an installation (like Homebrew) and shouldn't have to edit it directly.

freamon: Oh, I didn't think about that.  I've reverted it now

src/settings.rs
freamon: I've had another crack at it.  The behaviour now, is that 
'10' is in the code as a `Default`
`$MCFLY_RESULTS` over-rides this for general cases (e.g. user presses ctrl-r)
`--results` over-rides both for specific cases (user types out `mcfly search --results`)

cantino:
Thanks @freamon!

Issue #89, opened 2020-07-08
fish: how easy/hard would it be to add support for fish shell


mfrw:
Hi, i am wondering would it possible to make it work with fish shell.
I am planning to take a stab at it, if the author does not have any issue :)

zicklag:
If it makes any difference, I'd probably use this if it supported fish. :) :+1:

cantino:
I don't personally use fish but I'd gladly accept pull requests!

mfrw:
> I don't personally use fish but I'd gladly accept pull requests! Hopefully you can look at the bash and zsh integrations as a starting point, and I'm happy to answer questions.

Thank you very much :)
Will go through the bash/zsh impls

tjkirch:
I gave it a shot and have fish support working through a `mcfly.fish` file you source.  It doesn't allow you to remove commands from history yet, because the Rust code doesn't understand the fish history format, so I'll work on that next.

(Should #10 be closed since this issue has a little more recent traction?)

zicklag:
Sweet I'll test it out once the PR is up!

cantino:
Awesome! Yes, I'll close #10.

tjkirch:
Quick update.  I have fish history import working now as well, and everything's fine in fish 2.7 and 3.1.  History deletion is trickier, but I have an idea.

A little background is necessary.  fish has a structured history file (almost, but not quite, YAML) with entries that store command, time, and any relevant paths.  256K of these entries can be stored before the oldest start being purged; commands are automatically deduplicated.  This rotation/deduplication happens automatically every handful of commands.  (If you run ~10-20 distinct commands, you should see it by then.)

In short, it's a little more dynamic than bash/zsh, and it makes me uncomfortable reading, editing, and replacing the file.  You could lose history if fish did an update in the middle of mcfly's update.  I had initially implemented the naive method and saw surprising results during that cleanup.  (And the format is just more complex, so I was worried anyway.)

What you really want to do is have fish handle history deletion itself.  And we're already sending commands out to fish - in that respect, it's like `mcfly.zsh`, which saves `mcfly search` output to a temp file and interprets the results as commands.

[...]

PR #95, opened 2020-07-27
Add compatibility with fish shell

tjkirch:
Implementation notes:
* `mcfly.fish` is based on the bash/zsh implementations, with the key bindings function inspired by fzf.
  * Key bindings are added by running `mcfly_key_bindings` after sourcing the file, rather than automatically, in case the user doesn't like the default ones.
* The shebang is `/usr/bin/env fish` because I believe fish shell path is less likely to be consistent than bash/zsh.
* To simplify, usage with fish doesn't need a `MCFLY_HISTORY` file, because we can get the last command easily during a fish postexec function.  It's pointed at /dev/null, and temporary file cleanup is skipped.
* A new `HistoryFormat` enum, defaulting to the current format `Plain`, lets us specify fish-format history so we can import/load.
* A new `HistoryCommand` struct is used to represent each command in the history file, since it's no longer necessarily one line per entry.
* The format of `search -o` files was extended to allow sending more commands back to the shell, in particular a 'delete' command for fish, which needs to handle history deletion itself rather than through file edits.

Possible future improvements:
* There's talk of changing fish's history format in fish 3.2.  This could be handled with a new `FishJson` variant of `HistoryFormat` (named as appropriate, based on the format fish selects) and a `MCFLY_HISTORY_FORMAT` environment variable that the user can set to select it.
* Use zsh extended timestamp as the `when` in `HistoryCommand`, if available.
* Inspect zsh history file to see if extended timestamps are used, rather than requiring an argument, so that when the file is rewritten (when erasing lines) it doesn't lose extended timestamps.  (Note: upon exit, zsh rewrites the file again, restoring extended timestamps, so it's not the end of the world.)

Testing done:
* Tested add/search, importing history, and deleting history with bash 5.0.17, zsh 5.8, fish 2.7.1, and fish 3.1.2.

Fixes cantino/mcfly#89.


src/settings.rs
```diff
@@ -216,7 +243,12 @@ impl Settings {
                 );
 
                 settings.append_to_histfile = add_matches.is_present("append_to_histfile");
-                settings.zsh_extended_history = add_matches.is_present("zsh_extended_history");
+                if add_matches.is_present("zsh_extended_history") {
+                    match settings.history_format {
+                        HistoryFormat::Plain { .. } => settings.history_format = HistoryFormat::Plain { zsh_extended_history: true },
```
cantino, line 248: Yea, this would be cleaner if we just had three history formats, since zsh_extended_history really just implies zsh.

src/shell_history.rs
```diff
@@ -49,80 +52,161 @@ pub fn history_file_path() -> PathBuf {
...
-    let history_contents = read_ignoring_utf_errors(&path);
+/// Represents each entry in a history file.
+#[derive(Debug)]
+pub struct HistoryCommand {
+    /// The user's command.
+    pub command: String,
+    /// When the command was run, in seconds since Unix epoch.
+    pub when: i64,
+    /// The format of the file, so we can write the record back out.
+    pub format: HistoryFormat,
```
cantino, line 63: Do we need to store this on every HistoryCommand?

src/shell_history.rs
```diff
@@ -49,80 +52,161 @@ pub fn history_file_path() -> PathBuf {
...
+            let history_contents = read_ignoring_utf_errors(&path);
+            let zsh_timestamp_and_duration_regex = Regex::new(r"^: \d+:\d+;").unwrap();
+            let when = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_else(|err| panic!("McFly error: Time went backwards ({})", err))
+                .as_secs() as i64;
+            history_contents
+                .split('\n')
+                .filter(|line| !has_leading_timestamp(line) && !line.is_empty())
+                .map(|line| zsh_timestamp_and_duration_regex.replace(line, ""))
```
cantino, line 108: You've set this up nicely to actually use the ZSH timestamp if it exists, which we're currently throwing away. :+1:

src/main.rs
```diff
@@ -41,16 +48,29 @@ fn handle_search(settings: &Settings, history: &History) {
...
+            // Next, the desired commandline selected by the user.
+            out.push_str("commandline ");
             out.push_str(&cmd);
+            out.push('\n');
+
+            // Finally, any requests for deletion of commands from shell history, for cases where
+            // shells need to handle this natively instead of through us editing HISTFILE.
+            for delete_request in result.delete_requests {
+                out.push_str("delete ");
+                out.push_str(&delete_request);
```
cantino, line 71: What happens if it's a multi-line command?

src/main.rs
```diff
@@ -41,16 +48,29 @@ fn handle_search(settings: &Settings, history: &History) {
...
             if result.run {
-                out.push_str("run\n");
+                out.push_str("mode run\n");
             } else {
-                out.push_str("display\n");
+                out.push_str("mode display\n");
             }
 
+            // Next, the desired commandline selected by the user.
+            out.push_str("commandline ");
```
cantino, line 63: Do multi-line commands work?

mcfly.fish
```diff
@@ -0,0 +1,74 @@
...
+# Avoid loading this file more than once
+if test "$__MCFLY_LOADED" = "loaded"
+  exit 0
+end
+set -g __MCFLY_LOADED "loaded"
+
+# Note: we only use the history file for the session when this file was sourced.
+# Would have to reset this before calling mcfly if you want commands from another session later.
+if not set -q HISTFILE
+  set -gx HISTFILE (set -q XDG_DATA_HOME; and echo $XDG_DATA_HOME; or echo $HOME/.local/share)/fish/(set -q fish_history; and echo $fish_history; or echo fish)_history
```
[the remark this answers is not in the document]
tjkirch: Yeah, this is the documented logic for finding the history file.  `XDG_DATA_HOME` is the preferred directory prefix if set, and `fish_history` is used as the session name if it's set, otherwise it falls back to `~/.local/share` and `fish`, respectively.

https://fishshell.com/docs/current/cmds/history.html#customizing-the-name-of-the-history-file
https://fishshell.com/docs/current/index.html#history-search

mcfly.fish
```diff
@@ -0,0 +1,74 @@
...
+  set -g __MCFLY_OLD_PWD "$PWD"
+end
+
+function __mcfly_add_command -d 'Add run commands to McFly database' -e fish_postexec
+  # First, retain return code of last command before we lose it
+  set -l last_status $status
+  # Handle first call of this function after sourcing mcfly.fish, when the old PWD won't be set
+  set -q __MCFLY_OLD_PWD; or set -g __MCFLY_OLD_PWD "$PWD"
+
+  test -n "$MCFLY_DEBUG"; and echo mcfly.fish: Run eval $__MCFLY_CMD add --exit '$last_status' --old-dir '$__MCFLY_OLD_PWD' '$argv[1]'
```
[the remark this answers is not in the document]
tjkirch: This is a distinct difference with fish, yeah - there's less variable expansion / interpretation going on here.  If there are spaces, or special characters used by fish, etc. they're passed through to the command rather than interpreted, and don't need quoting.  I tested commands with spaces, subshells, etc. and they showed up in McFly history as I typed them, not interpreted.

https://fishshell.com/docs/current/tutorial.html#variables  (in particular, the "In bash" line clarifies the difference with respect to spaces.)

Note that the `eval` here is actually removing the quotes, so it's not the quotes saving us, just the difference described above.  (The `eval` is necessary for compatibility with fish 2.x; fish 3 lets you directly use expanded variables like `$__MCFLY_CMD` as commands.)

src/settings.rs
```diff
@@ -23,11 +23,20 @@ pub enum KeyScheme {
     Vim,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum HistoryFormat {
+    /// bash/zsh format with just a command in plain text.
+    Plain { zsh_extended_history: bool },
```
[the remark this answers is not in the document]
tjkirch: I can do this now if you think it's a good fit.  I hadn't for a couple silly reasons.  (1) The formats are effectively the same, given the handling McFly is already doing around the zsh timestamp.  (2) We couldn't assume "plain" as a default anymore, so we'd have to specify history format in all calls to McFly.  That's obviously easy to handle in `mcfly.bash` and `mcfly.zsh`, but I wasn't sure you'd want that to be a requirement for any other calls people might make to the binary?  It's probably fine...

src/shell_history.rs
tjkirch: We don't *have* to, but it allows us to write the command back out (to a history file, debug print, etc.) in its native format without having to pass a `HistoryFormat` around in more places, some of which didn't make sense to me.  We can't make the `Display` implementation take extra arguments, since we don't own the trait, for example.  I'm sure I could find a way to pass it around where needed if you like.

src/main.rs
tjkirch: It deletes part of it from from the McFly database, and I think nothing from fish history :/  fish stores multi-line commands on a single line, with "\n" where appropriate, so it's clearer than zsh and I think it can be made to work.  See next comment though...

src/main.rs
tjkirch: Not exactly.  I don't think they work currently for zsh either, so I thought it'd be better addressed in a separate change, what do you think?

fish does a bit better than zsh.  The full command gets stored in history.  It shows up in the interface when you have a matching search; it seems to indent later lines more than necessary, but it's all there.  When you select the command, though, you only get the first line.

zsh stores multi-line commands on multiple lines in the history file (latter lines with no extended timestamp) and we currently assume each line is one command, so you'd get multi-line commands broken (figuratively and literally) into multiple.  We could be a bit cheeky and assume that if we're using extended timestamps and the next line doesn't include one, it's part of the previous command... the alternative seems to be actually understanding what zsh considers a complete command, which might get ugly.

(Not sure about bash)

tjkirch:
^ This push rebases on master, per https://github.com/cantino/mcfly/pull/95#discussion_r461765585

src/settings.rs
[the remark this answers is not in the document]
cantino: If you'd rather wait, or leave it for me, that's fine too.

src/shell_history.rs
cantino: It's likely fine. I was wondering if it's use a lot more RAM, but it's probably fine.

src/main.rs
cantino: Yea, I think you're right. Multiline support is pretty broken across the board.

src/shell_history.rs
tjkirch: As it stands now, it's a small enum containing only a bool in the fields, and I confirmed it's only 1 byte.  I agree that it's a potential concern if we add more data inside the enum later, since it's stored for each historical command, but as-is it'd be dwarfed by the size of the commands themselves and their timestamps.  If history size overall becomes a problem, it seems possible to move to a more stream-oriented model, but it'd be a bit of an overhaul.  And if it's just a problem with extra fields in `HistoryFormat` added later, we'd have the alternative of moving `HistoryFormat` out of `HistoryCommand` at that point, too, at the cost of the bit of complexity I mentioned.

src/settings.rs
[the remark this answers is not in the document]
tjkirch: No problem, I'll try that now.

tjkirch:
^ This push addresses https://github.com/cantino/mcfly/pull/95#discussion_r461769704 by splitting `HistoryFormat::Plain` into `Bash` and `Zsh`, with `extended_history` only in `Zsh`.  (This might also help to handle bash's `HISTTIMEFORMAT` in the future, similar to zsh's extended history.)  It defaults to bash because that's probably most common, but the `mcfly.*sh` files all specify their format now.

scott2b:
While this brings up McFly history in fish for me with ctrl^R, I am unable to get it to actually execute anything. Hitting either Tab or Return simply exits from McFly as if I had hit Esc.

tjkirch:
> While this brings up McFly history in fish for me with ctrl^R, I am unable to get it to actually execute anything. Hitting either Tab or Return simply exits from McFly as if I had hit Esc.

@scott2b What version of fish are you running?

It might help to comment out the removal of the temp file in mcfly.fish, then try again, and paste the content of that file here. Then we can see if the problem is inside mcfly or in mcfly.fish.

scott2b:
I am using fish 3.1.2

My mcfly.output.... files consistently look like:

```
run
somecommand
```

Where somecommand is something that would have come from my history.

scott2b:
Oh, but there's one file that is different. It is named mcfly.XXXXXXXX.g83UFnZd, note the missing `.output.` part of the name.

And the contents look more like a full history (it's about 10000 lines). E.g. a sample:

```
: 1592274472:0;mv 05* 05.Week05
: 1592274492:0;mv 06* 06.Week06
: 1592274504:0;mv 07* 07.Week07
: 1592274534:0;git add Lectures/02.Week02
: 1592274540:0;git add Lectures/03.Week03
: 1592274546:0;git add Lectures/04.Week04
: 1592274550:0;git add Lectures/05.Week05
```

tjkirch:
> My mcfly.output.... files consistently look like:
> 
> ```
> run
> somecommand
> ```
> 
> Where somecommand is something that would have come from my history.

@scott2b That's the format of the McFly temp files *before* the change in this PR that makes them work with fish.  Are you sure you built and installed McFly with the change from this PR, and put it in your PATH?  The change isn't merged yet.

scott2b:
I manually copied mcfly.fish into my fish functions directory and am sourcing that file and calling the binding function in my config.fish file.

Are you saying I should do a full install of mcfly from this branch?

tjkirch:
> I manually copied mcfly.fish into my fish functions directory and am sourcing that file and calling the binding function in my config.fish file.
> 
> Are you saying I should do a full install of mcfly from this branch?

Yeah, the Rust code changed too, so you have to build from this branch and use that McFly binary.

scott2b:
Ok. My bad. I did not realize that. I should have dug deeper into the PR. Thank you for your time. I'll give that a shot.

scott2b:
That was it! This is really great. Thank you so much @tjkirch

mcfly.bash
```diff
@@ -19,8 +19,8 @@ fi
 # MCFLY_SESSION_ID is used by McFly internally to keep track of the commands from a particular terminal session.
 export MCFLY_SESSION_ID=$(dd if=/dev/urandom bs=256 count=1 2> /dev/null | env LC_ALL=C tr -dc 'a-zA-Z0-9' | head -c 24)
 
-# Find the binary
-MCFLY_PATH=${MCFLY_PATH:-$(which mcfly)}
+# Find the binary and apply default arguments
+MCFLY_PATH="${MCFLY_PATH:-$(which mcfly)} --history_format bash"
```
cantino, line 23: 5e3161d changes this to detect if `which mcfly` fails, so let's delete this default for now, since it defaults to bash anyway.

mcfly.bash
```diff
@@ -63,8 +63,8 @@ PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 #      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
   if set -o | grep "vi " | grep -q on; then
-    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search\C-j'"
+    bind "'\C-r': '\e0i#mcfly: \e\C-j ${MCFLY_PATH} search\C-j'"
```
cantino, line 66: Let's leave this as just `mcfly` for now for readability in the user's shell. I'm okay if we rely on the default being bash for now.

tjkirch:
^ This push addresses @cantino's concerns by rebasing on master, then removing the `MCFLY_PATH` modifications in `mcfly.bash` (so there are no longer any changes in that file) and incorporates the path check from the bash/zsh files into `mcfly.fish`.  Retested fish 2, fish 3, and bash OK.

cantino:
@tjkirch, I've been sometimes seeing `commandline` in my zsh output:

Screen Shot 2020-08-28 at 5 25 33 PM

tjkirch:
> @tjkirch, I've been sometimes seeing `commandline` in my zsh output:
> 
> Screen Shot 2020-08-28 at 5 25 33 PM

Interesting, I haven't seen that.  I can't think of anything in the code that would insert "commandline" literally.  I have a few thoughts:
* If you hit enter again when that happens, does it set the command line to the remainder of the line, allowing you to hit enter a second time to run it?  I'm not super familiar with fish internals around command execution, but it seems like it's just getting [here](https://github.com/cantino/mcfly/blob/4e3ca71d24433a2147c419967b5fbfc2c75a2d7d/mcfly.fish#L64) and dying rather than running it and setting the command line string.  (If you fake this scenario by just typing `commandline "echo hi"` it should be about the same; two enter presses will echo hi.)
* Related: do you only see that when using 'tab' to edit the command line, but never with 'enter' to execute the command line?
* What version of fish?
* If you're using fish 3.1.1, there was a fix related to 'eval' in 3.1.2, and while I don't think it's related, it'd be interesting to see if you still have the issue in 3.1.2, or if you can't upgrade, if you still have it after removing the 'eval' from mcfly.fish.
* It'd be helpful to see the mcfly_output file from a run where that happened, so we can make sure it had the expected commandline and mode lines.  Maybe set mcfly.fish to not delete the file temporarily?

cantino:
So I think this was user error on my part. I haven't seen it again. Thanks @tjkirch!

Issue #74, opened 2019-12-11
Implement fuzzy search


meridius:
It would be great if results contained not only commands exactly matching the input, but allow for imperfections.

For starters ignoring special characters (`[^a-zA-Z0-9]`). 
Another round could implement real fuzziness (chars in incorrect order, typos made by pressing neighboring keys, ...).

cantino:
I think that's a great idea. Search is in sqlite, so whatever we do needs to be implemented there. Are you interested in working on it?

meridius:
Sure, I could give it a try. 
From a quick scan of a code it seems I should just compute the fuzzy strings and add them as OR LIKE to  https://github.com/cantino/mcfly/blob/master/src/history/history.rs#L230 Could you confirm that please?

cantino:
Yes, that's correct.

infokiller:
I suggest looking at [skim](https://github.com/lotabout/skim)

cantino:
Right now you can put a `%` in your search to do semi-fuzzy searching, but skim looks very nice. I'd be open to contributions that use it.

PR #102, opened 2020-09-26
--fuzzy search entire history with skim

dmfay:
In #74:

> Right now you can put a % in your search to do semi-fuzzy searching, but skim looks very nice. I'd be open to contributions that use it.

Here's a minimal skim implementation, enabled with a `--fuzzy` option in `mcfly search`. Near future considerations:

* options are all hardcoded at the moment
* skim either selects or doesn't, so there's no distinguishing Enter-to-run vs Tab-to-stage
* skim has its own UI and event loop, which can't display at the same time as mcfly's

It's a lot to give up for fuzzy search! I think it's absolutely worth it (I found mcfly a couple days after I wrote [my own non-neural but very fuzzy history searcher](https://gitlab.com/dmfay/mnem)) but I don't know how better it might be integrated than with `--fuzzy` essentially turning mcfly into a different application.


src/interface.rs
```diff
@@ -182,6 +190,39 @@ impl<'a> Interface<'a> {
         screen.flush().unwrap();
     }
 
+    fn skim(&mut self) {
+        self.selection = 0;
+        self.matches = self
+            .history
+            .find_matches(&self.input.command, -1);
```
cantino, line 197: How slow is this when you have a large command history?

cantino:
Cool @dmfay, thank you for contributing and welcome to the project!

src/interface.rs
dmfay: Thanks to an errant config setting I only have a couple thousand lines of real history, but I whipped up a [script](https://gist.github.com/dmfay/32a327869c40ec2673d622957a7b6289) to seed more. Time is me hitting Esc as soon as skim loads.

```
wc -l .histfile
101838 .histfile

time mcfly search --fuzzy
mcfly search --fuzzy  3.36s user 0.17s system 99% cpu 3.543 total
```

After that three-second load (😬), skim's performance is quite acceptable. I can type just faster than it can process 90-100k elements, but it buffers input and responds without stuttering or glitching.

I did notice some drag after _other_ commands as long as I kept the 100k-line history file -- a bit under a second in between the command completing and my prompt printing again. It went away upon replacing my original .histfile so I think that might be the shell, not mcfly.

cantino: So, I think we need to find something faster. I currently have 50000 commands in my mcfly history and it launches in ~200ms.

Maybe we can only start searching once you've typed a few characters?

dmfay: I was running with debug symbols! :woman_facepalming:

Built for release it's sub-second, in the 850-920ms range, which is hardly instant but still a far cry from three seconds. The version I installed from my package manager loads the same history in 600-650ms.

cantino: I'd be open to adding it now that I know it's sub-second and since it's optional. Maybe we should have it also be enable-able with an environment variable so that users can add that to their shell config files before sourcing mcfly?

I'd also love to see how your "%s everywhere" solution does, it might be good enough.

dmfay: The "%s everywhere" solution only does anything different starting on the first keystroke, so the initial load time is the same. At 90-100k elements, it's about a quarter-second faster than this. Performance after load is comparable if not even more responsive.

I've set up a second pull request with that version.

rkachach:
Hi there! thank you very much for this great tool. I'm trying to enable this feature setting the env variable:

`export MCFLY_FUZZY=true
`

 but it seems not available in my version (0.5.0) I'm doing something wrong or do I need the latest version to use this feature?

dmfay:
It is not quite available yet. This pull request was an earlier experiment that didn't quite pan out, but [this implementation](https://github.com/cantino/mcfly/pull/103) is awaiting merge. You can build it from source if you clone `git@github.com:dmfay/mcfly` and check out the `sqlite-fuzzy` branch, or wait for Andrew to merge and release.

rkachach:
Understood. Thanks for the quick response. I'd rather prefer to have it merged. I use fuzzy searching quite intensively in Emacs that's why I asked for the feature :+1:. Thank you very much for this great contribution.

cantino:
Released in v0.5.1 :)

PR #104, opened 2020-10-10
use c-m instead of c-j for return key

Konfekt:
The latter key is less likely to be rebound.

To avoid all these binding conflicts, one could think about a function `f` and a binding `bind -x "'\C-r': f"`instead

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -67,8 +67,8 @@ PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 #      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
   if set -o | grep "vi " | grep -q on; then
-    bind "'\C-r': '\e0i#mcfly: \e\C-j mcfly search\C-j'"
+    bind "'\C-r': '\e0i#mcfly: \e\C-m mcfly search\C-m'"
   else
-    bind "'\C-r': '\C-amcfly: \e# mcfly search\C-j'"
+    bind "'\C-r': '\C-amcfly: \e# mcfly search\C-m'"
   fi
 fi


cantino:
Thanks @Konfekt! This seems to work fine for me. Can you explain what the difference between c-m and c-j are, conceptually? Will this break anything for folks?

Konfekt:
In Vim c-m is mapped to `CR`(`\n`) whereas c-j is mapped to `NL` (`\r`).
In `.inputrc` one can bind `c-j` to a command different from that of `CR`, but not `c-m`
According to https://stackoverflow.com/a/14619214

> Enter is usually equivalent to C-m. But, if the icrnl flag is active for the tty (see stty -a), then an input C-m will automatically be translated to C-j.

Therefore, mapping it to c-m is likely to cause fewer collisions. 

As an alternative independent of key bindings,  there is
 
```sh
f() { mcfly search; }
bind -x "'\C-r': f"
```

where probably [bind -x](https://www.gnu.org/software/bash/manual/bash.html#Bash-Builtins) has been around for a while, at least since 2014.

cantino:
Thanks for the explanation! Let's start with this and then we can try to move to `bind -x` to fix `#mcfly:` showing up in the history. See #105

Konfekt:
Somehow `mcfly` now (at least since version >= 5.12) always maps `Ctrl-J` to `Ctrl-M`; it might be a change in a terminal input library?

cantino:
Locally I just did ctrl-j and it accepted the selection. What did it do before?

Konfekt:
That depended on the value of "\C-j" in `~/.inputrc`; if not set to `accept-line` it would not accept the selection

cantino:
I don't think that's true. It's explicitly bound to accept the selection,
or am I missing something?
https://github.com/cantino/mcfly/blob/874441b6bcdc5292375da654fd131146f547a8a4/src/interface.rs#L518

cantino:
No worries! I'd be open to someone upgrading McFly to use .inputrc.

Issue #100, opened 2020-08-28
When exiting (ESC) from mcfly search (or Ctrl+R) I see unwanted output


asdf23:
What I'm doing:
Ctrl + R > ESC
or `myfly search` > ESC

I expect the screen to return to how it looked before my last command. Instead I see:
```[user1@myBox ~]$ touch something # <--- last command before hitting Ctrl+R then ESC
[user1@myBox ~]$ #mcfly: 
[user1@myBox ~]$  mcfly search
[user1@myBox ~]$ 
```

cantino:
As far as I know, this is unavoidable in bash. (Solutions welcome!) It doesn't happen in zsh or fish.

CreativeCactus:
This is certainly an issue. Please reopen.

The problem is described [here](https://github.com/cantino/mcfly/blob/5e3161d6a807596faa704f8528b349691d947b41/mcfly.bash#L60-L74):

- Get the contents of the current line, to be used in the search query
- Invoke `mcfly search (query)` or insert the search query in the history file and invoke `mcfly search`

Bonus:

- When cancelled, leave the terminal with the old current line, or contents of the search query

CreativeCactus:
**Nothing is unavoidable or impossible, it is only a matter of practicality for the hacks involved.** In this case, it seems like it's only a small change.

With about 40 mins of searching and testing, I found that you can invoke functions using `bind`, is there some reason that this feature isn't being used? I couldn't find a bash version where `bind` was introduced, but I assume it is quite old, and in case we need to check for support, what's another nested `if` statement between friends?

By replacing [this line](https://github.com/cantino/mcfly/blob/5e3161d6a807596faa704f8528b349691d947b41/mcfly.bash#L70) with `bind -x '"\C-r": "array=($READLINE_LINE); echo \#mcfly: ${array[0]} > $MCFLY_HISTORY ; mcfly search"'` I was able to get the exact functionality described above.

I have made a [PR here](https://github.com/cantino/mcfly/pull/105).

PR #105, opened 2020-10-15
feat/ Fix for issue 100, removes terminal debris

CreativeCactus:
Fixes https://github.com/cantino/mcfly/issues/100

I don't know enough about the first `bind` if `set -o | grep "vi " | grep -q on` passes, but I'm sure someone who knows why this condition exists can copy my pattern. :)

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -27,7 +27,7 @@ if [ -z "$MCFLY_PATH" ]; then
 fi
 
 # Ignore commands with a leading space
-export HISTCONTROL="ignorespace"
+export HISTCONTROL="${HISTCONTROL:-ignorespace}"
 
 # Append new history items to .bash_history
 shopt -s histappend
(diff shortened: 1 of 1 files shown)


cantino:
Thanks @CreativeCactus, this looks like a great start to a solution! However, on my computer, it echos `#mcfly:` to the history and never includes the current command line. It looks like `array[0]` is always blank for me in bash (3.2) on OS X.

CreativeCactus:
It looks like [bash <4 has no `READLINE_LINE`](https://unix.stackexchange.com/a/106832), so we will either need a version check, or find another way to get the current line in the bind handler.

Also, `${array[0]}` should have been `${array[@]}`, my bad.

CreativeCactus:
@cantino 
I've added a version check to use the `$READLINE_LINE` trick.

For `bash<4`, you might be able to use those control characters to move up a line and clear it? It's not as elegant, but it might be worth trying if you have not already. Maybe even `bind -x '"C-r": "printf some control characters"`.

Unfortunately I don't have much time to experiment this week, but this branch should now be suitable to merge (pending review) as a fix for `bash>=4` users.

cantino:
Sorry for the delay on this, I'll need to test it out on a virtual machine when I have a chance. It's too bad that it needs to version check bash, but I guess that's unavoidable.

cantino:
I've rebased and updated this in #113

PR #113, opened 2020-12-06
Prevent cluttering up the buffer in Bash 4+

cantino:
Thanks to @CreativeCactus for the methodology. I think this works, but it'd be helpful if users of bash could test it. I use zsh these days.

Things to test:
- Typing and hitting ctrl-r searches for what you typed, and selecting something now should clear the buffer, not append to what you typed before.
- `vi` mode works. (I think this only works when in edit mode.)

@JamJar00

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -58,17 +58,21 @@ function mcfly_prompt_command {
 PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
 
 # If this is an interactive shell, take ownership of ctrl-r.
-# The logic here is:
-#   1. Jump to the beginning of the edit buffer, add 'mcfly: ', and comment out the current line. We comment out the line
-#      to ensure that all possible special characters, including backticks, are ignored. This commented out line will
-#      end up as the most recent entry in the $MCFLY_HISTORY file.
-#   2. Type "mcfly search" and then run the command. McFly will pull the last line from the $MCFLY_HISTORY file,
-#      which should be the commented-out search from step #1. It will then remove that line from the history file and
-#      render the search UI pre-filled with it.
 if [[ $- =~ .*i.* ]]; then
-  if set -o | grep "vi " | grep -q on; then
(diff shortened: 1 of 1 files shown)


JamJar00:
I can probably have a look, I think I've deleted my linux VM with all the rust build tools on though!

JamJar00:
Tested in WSL 1 running bash `4.4.19(1)-release` and this all seems to be working.

I can theoretically test on an Ubuntu VM as well but it'll take me a little longer to install everything!

JamJar00:
For clarity, it's still outputting:
```
$ #mcfly: whatever you type
$  mcfly search
```
But I don't think the PR is trying to solve that.

Neither of those lines are in the bash history, just on the console.

CreativeCactus (approved):
This works for me on Bash 4.4.23

CreativeCactus:
> For clarity, it's still outputting:
> 
> ```
> $ #mcfly: whatever you type
> $  mcfly search
> ```
> 
> But I don't think the PR is trying to solve that.

This PR should solve that. It might be that you are using bash <4, or some other shell. The `bind -x` approach writes `#mcfly: ` directly to the `$MCFLY_HISTORY` file for compatibility, but really it could use any mechanism to pass history to mcfly. Doing it this way *does* prevent writing to the terminal, and therefore does not depend on `$HISTCONTROL` to hide it from bash history.

JamJar00:
> This PR should solve that. It might be that you are using bash <4, or some other shell. The `bind -x` approach writes `#mcfly: ` directly to the `$MCFLY_HISTORY` file for compatibility, but really it could use any mechanism to pass history to mcfly. Doing it this way _does_ prevent writing to the terminal, and therefore does not depend on `$HISTCONTROL` to hide it from bash history.

You are absolutely correct, I forgot to source in my bashrc after I built it, looks like I just tested normal mcfly 🤦‍♂️ 

Everything seems to be working alright for me!

JamJar00 (approved):

cantino:
Thanks for testing folks!

cantino:
Released in 0.5.2

Issue #121, opened 2021-01-27
Consider moving $HOME/.mcfly directory to $HOME/.config/mcfly in compliance with the FreeDesktop XDG folders specification


Awernx:
Hello,
Thanks for building this awesome tool! 
To avoid cluttering the HOME directory, a lot of modern applications have started moving their configuration directory to be under $HOME/.config in compliance with the FreeDesktop XDG folders specification. 

More information on this standard is available at https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

Thanks!

cantino:
Thanks @Awernx! I'd be open to a pull request that allows setting of the config dir via environment variable. We'd need to have mcfly continue to at least check for the existence of `~/.mcfly` by default though.

Konfekt:
As the database stores data rather than settings, `$XDG_DATA_HOME` seems more appropriate.

However, since `.mcfly/history.db` depends on `$HISTFILE`, most convenient would be to point `$HISTFILE` to a folder, say `$XDG_DATA_HOME/sh`, and store `.mcfly/history.db` in it, too.

Awernx:
@Konfekt  - Agreed. I have submitted a pull request that uses XDG_DATA_HOME for history.db, and XDG_CACHE_HOME for mcfly_training_cache_path

Konfekt:
Good work! Just wanted to note that having $HISTFILE and history.db in the same folder makes porting and protecting the shell history easier.

judaew:
I would be happy if there was an environment variable that could set the necessary directory. Like to `pass`, `bundle` or `cargo`:

```bash
export PASSWORD_STORE_DIR=$HOME/.local/share/pass

export BUNDLE_USER_CONFIG=$HOME/.config/bundle
export BUNDLE_USER_CACHE=$HOME/.cache/bundle
export BUNDLE_USER_PLUGIN=$HOME/.local/share/bundle

export CARGO_HOME=$HOME/cargo
```

PR #124, opened 2021-02-07
Storing configuration in XDG directories, if ~/.mcfly does not exist

Awernx:
Hello @cantino! Per your suggestion on #121, I've created a pull request.

Code checks for existence of ~/.mcfly, and uses it if available. Otherwise, 
1. **history.db** is stored in XDG_DATA_DIR
2. **cache CSV** is stored in XDG_CACHE_DIR

Code uses [directories-next](https://crates.io/crates/directories-next) for XDG path computation.

Thanks!

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -400,7 +400,7 @@ impl Settings {
     }
     
     fn mcfly_xdg_dir() -> ProjectDirs {
-        ProjectDirs::from("io", "github",  "McFly").unwrap()
+        ProjectDirs::from("", "",  "McFly").unwrap()
     }
     
     fn mcfly_base_path(base_dir: PathBuf) -> PathBuf {
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -2,7 +2,7 @@ use crate::shell_history;
 use clap::AppSettings;
 use clap::{crate_authors, crate_version, value_t};
 use clap::{App, Arg, SubCommand};
-use dirs::home_dir;
+use directories_next::{UserDirs, ProjectDirs};
 use std::env;
 use std::path::PathBuf;
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


src/settings.rs
```diff
@@ -385,17 +385,41 @@ impl Settings {
...
+    // Use ~/.mcfly only if it already exists, otherwise create 'mcfly' folder in XDG_DATA_DIR 
     pub fn mcfly_db_path() -> PathBuf {
-        Settings::storage_dir_path().join(PathBuf::from("history.db"))
+        let data_dir = Settings::mcfly_xdg_dir().data_dir().to_path_buf();
+    
+        Settings::mcfly_base_path(data_dir).join(PathBuf::from("history.db"))
+    }
+    
+    fn mcfly_xdg_dir() -> ProjectDirs {
+        ProjectDirs::from("io", "github",  "McFly").unwrap()
```
cantino, line 403: What is the recommend standard for this?

src/settings.rs
```diff
@@ -385,17 +385,41 @@ impl Settings {
...
+    }
+    
+    fn mcfly_base_path(base_dir: PathBuf) -> PathBuf {
+        let (home_exists, mcfly_path) = Settings::home_mcfly_exists();
+    
+        if home_exists {
+            mcfly_path
+        }
```
cantino, line 411: Please run `cargo clippy` and `cargo fmt` on this commit, I think the formatting is slightly off.

cantino (commented):
Do you think we should migrate users' files?

Could you also update the README to document where the files will be created?

src/settings.rs
jirutka: IIRC I’ve never seen such convention. `~/.local/share/mcfly` and `~/.cache/mcfly` would be fine and more standard.

Awernx: Just to clarify - there isn't a specific convention on the 3 three fields - qualifier, organization and project. As a matter of fact, qualifier and org names affect the folder names on MacOS and Windows, but not Linux. **That said these 2 fields can be empty, and I'll make them so**. In the long term, depending on the direction of this project, these fields can be updated.

Additional information and guidance on this is available [here](https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.from)

src/settings.rs
Awernx: Done

Awernx:
Have made the suggested changes

colemickens:
This change looks good, I'd like to see it merged. `mcfly` currently stands aside ssh+gnupg+parallel as polluting my homedir and I am hoping to knock this one off the list. Thanks for mcfly and the change!

cantino:
`~/.mcfly` feels like a good default to me, but maybe I'm old school? I'd definitely be fine with this being configurable with a setting, but I don't know how I feel about moving the default.

nfultz:
> `~/.mcfly` feels like a good default to me, but maybe I'm old school? I'd definitely be fine with this being configurable with a setting, but I don't know how I feel about moving the default.

If you are thinking of adding an environment variable to configure it, you may as well just piggyback off of  $XDG_CONFIG_HOME - which is what this PR does. 

Or continue to use `~/.mcfly`, which is also still supported.

colemickens:
Respectfully, this feels like a safe change, especially after reviewing the code:
* this won't affect existing users
* still allows for `~/.mcfly`
* respects the XDG desktop guidelines, and I'm desperate for Linux on desktop to have some semblance of sanity and care

*Nearly every* other greenfield Rust app follow xdg - `bottom`, `helix`, `zellij`, `wezterm`, `alacritty`, `jj`, `joshuto`, `nu`, `pijul`, etc... all respect XDG dirs. `mcfly` stands alone here; even `zoxide` follows it.

And, as of now, `.mcfly` and `.mozilla` are now the only thing left in my home dir, beyond the unmovable `.ssh`:

```console
~
❯ exa -la
drwxr-xr-x  - cole  8 Mar 14:32 .cache
drwx------  - cole  8 Mar 14:26 .config
drwxr-xr-x  - cole 26 Nov  2021 .local
drwxr-xr-x  - cole  8 Mar 14:33 .mcfly
drwx------  - cole 26 Nov  2021 .mozilla
drwxr-xr-x  - cole  8 Mar 14:26 .ssh
drwxr-xr-x  - cole  7 Mar 01:09 code
drwxr-xr-x  - cole  8 Mar 13:58 downloads
drwxr-xr-x  - cole  8 Mar 14:10 Sync
lrwxrwxrwx 70 cole  8 Mar 14:26 .zshenv -> /nix/store/vzsbwrcq46mddx1ra29pwqlfzf2k9vm5-home-manager-files/.zshenv
```

Even Mozilla is finally fixing this after a 18+ year old open bug: https://bugzilla.mozilla.org/show_bug.cgi?id=259356

I'm not sure I consider `mcfly` special enough to be one of the last things dumping into my home directory.

cantino:
To be replaced by https://github.com/cantino/mcfly/pull/244

cantino:
https://github.com/cantino/mcfly/pull/244 has been merged

colemickens:
Awesome, thank you @cantino and @Awernx

cantino:
:+1: I'll get a new build out shortly.

cantino:
https://github.com/cantino/mcfly/releases/tag/v0.6.0 released!

PR #132, opened 2021-03-13
Fix #130 by using defult values `0` when table is empty.

### LINKED ISSUE #130: Deleting all items causes unwrap panic
This is an edge case from a new machine with no commands (!) yet run, but also reproducible if you F2 delete every command... upon pressing ctrl+R:
```
❱ thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidColumnType(0, Null)', /Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/rusqlite-0.15.0/src/row.rs:149:31
```
Once I run a command then ctrl+R runs without panic.

expectation: either errors with a message saying "no commands in history" or show the empty list.

akinnane:
Wrapping the aggregate functions with `IFNULL` allows `query_row` to return successfully preventing the unwrap error from rusqlite. `0` might not be the best default value but it seems to work.

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -446,7 +446,7 @@ impl History {
         let (mut when_run_min, when_run_max): (f64, f64) = self
             .connection
             .query_row(
-                "SELECT MIN(when_run), MAX(when_run) FROM commands",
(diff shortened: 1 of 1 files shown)


cantino:
Thanks!

PR #134, opened 2021-03-14
Use isatty in fish shell

domoritz:
See https://fishshell.com/docs/current/cmds/isatty.html


cantino:
Does this fix #129? My interpretation of what you said in that issue was that it does not.

domoritz:
You're right, it does not but it is an improvement nonetheless since isatty is more robust in general.

cantino:
Okay!

PR #137, opened 2021-04-01
Fix fish initialization

### LINKED ISSUE #129: Unknown command: mcfly_key_bindings on fish shell
I installed McFly via home-brew. I have version McFly 0.5.4. I also added 

```
# mcfly

mcfly init fish | source
mcfly_key_bindings

set -gx MCFLY_LIGHT true
set -gx MCFLY_FUZZY true
```

to the end of my `config.fish` file. I have fish version 3.2.0.  

Even when I run `mcfly init fish | source` in my terminal, `mcfly_key_bindings` is still not defined.

domoritz:
Fixes #129

Big thanks to @zanchey for helping me make sense of isatty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ When suggesting a command, McFly takes into consideration:
     Fish:
     ```bash
     mcfly init fish | source
-    mcfly_key_bindings
     ```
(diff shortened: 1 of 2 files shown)


cantino:
Released in v0.5.6

PR #138, opened 2021-04-01
fix: don't interpret search contents

scooter-dangle:
In the fish script, calling

```fish
 # Note: mcfly opts omitted for brevity
mcfly search (commandline)
```

can cause issues if the command line contained anything that mcfly
parses as an option. E.g., if I wanted to search for the last time I ran
a command that a I know included `-o 5`, the interpolated mcfly call
would be

```fish
mcfly search -o 5
```

which causes issues since `mcfly search` will assume `-o 5` is
specifying an output selection of `5`.

- - -

Note: I explicitly verified the bug when calling CTRL-R with `-o 5` on the command line. I didn't look into how `__mcfly_add_command` was being called. Maybe the `--` precaution isn't needed there?


cantino:
Thank you!

cantino:
Released in v0.5.6

PR #141, opened 2021-04-08
Don't create a tarbomb

barmic:
tarbomb (archive without folder) is generaly a bad practice.

Instead of:

- mcfly
- mcfly.bash
- mcfly.fish
- mcfly.zsh

this will create:

- mycfly
  - mcfly
  - mcfly.bash
  - mcfly.fish
  - mcfly.zsh

I'm not sure of all impact of it (for hombrew for example).

### DIFF
--- a/ci/before_deploy.sh
+++ b/ci/before_deploy.sh
@@ -26,8 +26,7 @@ main() {
     cp mcfly.zsh $stage/
     cp mcfly.fish $stage/
 
-    cd $stage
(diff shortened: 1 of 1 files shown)


cantino:
Thanks @barmic! However, I think this might break Homebrew and users' install scripts.

necaris:
@barmic if you still have the energy to look at it, it would be great if this PR also included the changes to the Homebrew recipe and install documentation to go with the changed folder structure in the zip file.

PR #145, opened 2021-05-10
Minor lint fixes

SafariMonkey:
Fixes some Rust warnings:
* `panic!(format!(...))` is redundant, as `panic!` supports formatting
* `mem::replace` expects its return value to be used, but it's dropped - I changed this to an assignment, but I'm not 100% sure if this is equivalent

Also fixes some "issues" in the Bash scripts, mostly reported by ShellCheck:
* it's good practice to quote variables, so I quoted the ones ShellCheck complained about
* there are some places where it complained about the exit code being suppressed, so I propagated it
  * it's not clear to me whether this is at all useful to you, as `set -e` and such isn't being used

### DIFF
--- a/dev.bash
+++ b/dev.bash
@@ -1,7 +1,9 @@
 #!/bin/bash
 # Build mcfly and run a dev environment bash for local mcfly testing
 
-this_dir=$(cd `dirname "$0"`; pwd)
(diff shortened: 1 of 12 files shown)


cantino:
Thanks @SafariMonkey!

PR #146, opened 2021-05-10
Initialize the database inside a transaction

SafariMonkey:
I tried initializing `mcfly` on my ~1.5MB history file, and after leaving it for several minutes it hadn't finished initializing. Apparently, [sqlite is known to be slow when inserting a lot of records outside of a transaction](https://stackoverflow.com/q/3852068). As such, I tried adding a transaction around the operation, and sure enough, it initialized in a second or two.


cantino:
Tested it and it seems to work. Nice trick!

PR #149, opened 2021-05-23
Show last run time for each command

dmfay:
with thanks to [atuin](https://github.com/ellie/atuin) for the idea of showing timing!

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -245,27 +245,13 @@ impl History {
 
         like_query.push_str("%");
 
-        let query = "WITH matched_commands AS (
-                           SELECT id, cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
-                                  age_factor, length_factor, exit_factor, recent_failure_factor,
-                                  selected_dir_factor, dir_factor, overlap_factor, immediate_overlap_factor,
-                                  selected_occurrences_factor, occurrences_factor
-                           FROM contextual_commands
-                           WHERE cmd LIKE (:like)
-                           ORDER BY rank DESC LIMIT :limit
-                     ), by_cmd AS (
-                           SELECT cmd, MAX(when_run) AS last_run
-                           FROM commands
-                           WHERE cmd IN (SELECT cmd FROM matched_commands)
-                           GROUP BY cmd
-                     )
-                     SELECT id, matched_commands.cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
-                            age_factor, length_factor, exit_factor, recent_failure_factor,
-                            selected_dir_factor, dir_factor, overlap_factor, immediate_overlap_factor,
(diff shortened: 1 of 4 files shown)


src/history/history.rs
```diff
@@ -244,13 +245,27 @@ impl History {
 
         like_query.push_str("%");
 
-        let query = "SELECT id, cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
+        let query = "WITH matched_commands AS (
+                           SELECT id, cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
```
cantino, line 249: Does using this `when_run` not work, such that you need to do the second query?  If this isn't the most recent, maybe  when the contextual_commands temp table is made, it should be `MAX(when_run)`?

cantino:
Could we make it use less space by just showing `5d` or `30m` or `10s`. I don't think `30m 10s` adds much?

src/history/history.rs
dmfay: There would need to be at least two queries here to get the max grouped only by `cmd` but on looking temp table creation is a _much_ better place for this!

dmfay:
I've liked having days-hours and hours-minutes resolution in my test-driving so far, so I may have sold it a bit short with the recent git commands. That said, other scales are definitely not as useful, so that specific band really has to be worth the four characters. Try it out for a day or three and see what you think, maybe?

cantino:
> I've liked having days-hours and hours-minutes resolution in my test-driving so far, so I may have sold it a bit short with the recent git commands. That said, other scales are definitely not as useful, so that specific band really has to be worth the four characters. Try it out for a day or three and see what you think, maybe?

Ok!

cantino:
I don't mind the two part dates, but I am finding it a bit hard to parse when I skim the list. I wonder if we should make it a lighter color (on dark backgrounds) and put it flush to the right?

dmfay:
here's a version with timings on the right. The FixedLengthGraphemeString looks like it only checks on input and returns whatever the internal string has accumulated (even if shorter), though, and I'm not quite sure what to do about that. We have `width` but do we have the current cursor x somewhere?

cantino:
I'm not 100% sure what you're asking. If you need to know if the line is currently selected in `truncate_for_display`, you could pass in `index == self.selection` to it. Is that what you're asking? All FixedLengthGraphemeString does is ensure that you cannot add more than N unicode characters to the string.

Do you like how it looks if you put 

```
write!(screen, "{}", cursor::Goto(width - 9, index as u16 + RESULTS_TOP_INDEX)).unwrap();
```

on line 248?

dmfay:
I was trying to fill the remaining width between command text and timing info with spaces and not having much luck. `cursor::Goto` does the trick, thanks!

cantino:
Thanks @dmfay, this is a nice addition :)

Issue #150, opened 2021-05-23
Implement way to automatically detect dark/light terminal


cmer:
I'd like to propose the following:

If `MCFLY_LIGHT` is set to `TRUE` or `FALSE`, force McFly to display in the specified mode.

If `MCFLY_LIGHT` is NOT set, make a best effort guess to determine which mode to display in, otherwise fall back to dark mode. I was able to detect light/dark mode in some terminals (but not all) with the following Fish script:

```
  if string match -r '\;15$' $COLORFGBG &> /dev/null
    set -gx MCFLY_LIGHT TRUE
  end
```

It'd be awesome if McFly could perform this check every time it is triggered when `MCFLY_LIGHT` is not set to any particular value.

The `COLORFGBG` environment variable is set by many terminals such as iTerm2, Rxvt and Konsole. macOS Terminal does NOT set it, however.

jltml:
Totally agreed; this would be wonderful! For anyone stumbling upon this who uses zsh on Mac with Terminal.app, this is what I have in the interim:

```zsh
mcfly_auto_light_mode() {
  if [[ $(defaults read -g AppleInterfaceStyle) != 'Dark' ]]; then
    export MCFLY_LIGHT=TRUE
  fi
}
zsh-defer mcfly_auto_light_mode
```

(the `zsh-defer` is from [this plugin](https://github.com/romkatv/zsh-defer))

cantino:
I'm game if you want to make this change @cmer. I think the determination should be in the mcfly.bash / mcfly.zsh / mcfly.fish scripts that then set the ENV.

cmer:
I have submitted a PR.

WladyX:
This would apply only to new terminal/tabs right? ie it would not apply for existing ones.
I have hacked my .p10k so this gets set on each prompt iteration:
```
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS+=(mcfly_theme)
function prompt_mcfly_theme() {
  if grep -q 'light' ~/.theme_status; then
    export MCFLY_LIGHT=TRUE
  else unset MCFLY_LIGHT
  fi
}```
would love for mcfly to have a config where i could set this (so it checks for the theme on each run) and i could update that file with https://github.com/cormacrelf/dark-notify

Tealk:
For Gnome 43 it works like this:
```bash
colorscheme="$(gsettings get org.gnome.desktop.interface color-scheme)"
[...]

PR #151, opened 2021-05-24
Automatically detect light terminal on initialization. Implements #150.

cmer:


cantino:
Does this need something like from @jltml's [comment](https://github.com/cantino/mcfly/issues/150#issuecomment-846626586)?

```zsh
  if [[ $(defaults read -g AppleInterfaceStyle) != 'Dark' ]]; then
    export MCFLY_LIGHT=TRUE
  fi
```

cmer:
I also implemented Apple Terminal light mode detection. 

One thing to keep in mind, though, is that this method sets the color scheme when the shell is first initialized. If dark mode changes after the shell has been initialized, Mcfly will display the wrong color scheme. This is partly why I had first suggested implementing this logic directly in the Rust codebase so that the checks can be made on each invocation of Mcfly.

This situation can happen if people's color scheme changes automatically at day/night time, for example.

Would you consider implementing this in the main codebase, @cantino?

cantino:
That's a good point. I guess it's a bit annoying to have to call things like `defaults read -g AppleInterfaceStyle` from within the Rust code, but maybe it's okay. What do you think?

jltml:
Totally up to you — I think it would be really nice to have it within the Rust codebase so it's dynamic/set upon each invocation (I have a tendency to leave Terminal windows open for days, so this would be really nice), but I agree with the annoyingness of having it in the Rust codebase too. Also, I just timed how long `defaults read -g AppleInterfaceStyle` takes with [Hyperfine](https://github.com/sharkdp/hyperfine), and on my 2017 i5 MacBook Pro it took ≈ 8.5 ms ± 0.8 ms… so I suppose it would make startup that little bit slower too. It's definitely a trade-off, but maybe it could be controlled by an environment variable — something like `MCFLY_LIGHT=auto`? — so that there's a choice between saving the 8ms and setting the theme dynamically? Just an idea; it could very well be a terrible one :)

I'd love to implement this myself, but unfortunately I have yet to learn Rust!

cmer:
It'd be great if we could only rely on env variables but it's not the case. However, that slight slow down would only happen on Apple Terminal, and if `MCFLY_LIGHT` is not set. It might be worth the tradeoff.

I suggest:

- treating `MCFLY_LIGHT` as "auto" when it is not set, 
- forcing light mode when set to `TRUE`
- forcing dark mode when set to `FALSE`
 
The "magic" would only happen when `MCFLY_LIGHT` is not set at all. This holds true for all terminals, not just Apple Terminal.

cantino:
> slight slow down would only happen on Apple Terminal, and if `MCFLY_LIGHT` is not set

How do we know when we're in that terminal and need to shell out to `defaults`?

cmer:
See my implementation here: https://github.com/cantino/mcfly/pull/151/commits/55554beb8fd94a12edb3c28ca9ff682b87f59ef8

cantino:
What do you think about this in relation to #156?

cmer:
In my opinion, Mcfly should always try to do the right thing automatically without the need for any configuration, which is why I think it'd be wise to have something directly in Mcfly to try to automatically detect light mode, especially since it's rather trivial to do it effectively in 95% of cases.

As an example, I don't need to configure `ls` for it to work. It just does. Just my 2 cents!

cantino:
I think having this in the shell scripts is simpler for the time being, but eventually could be pulled into the Rust code. Is this PR ready to merge? (@dmfay & @SafariMonkey appreciate a review on this too if you can.)

SafariMonkey:
I would suggest keeping the previous value (if set) using something like `export MCFLY_LIGHT="${MCFLY_LIGHT:-TRUE}"` ([`bash`](https://wiki.bash-hackers.org/syntax/pe#use_a_default_value) , [`zsh`](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion)), which seems to be [roughly equivalent to](https://stackoverflow.com/a/54524966/7177899) changing to `test -z MCFLY_LIGHT; and set -gx MCFLY_LIGHT TRUE` in `fish`.

Otherwise, it looks fine to me, but I'm far from an expert.

AllanOricil:
Seems that it does not work on the latest macos

image

AllanOricil:
there should exist a standard for all OS to return if it is using dark/light mode

PR #160, opened 2021-06-18
Clean PROMPT_COMMAND before joining with a semicolon

### LINKED ISSUE #159: Conflict with zoxide
```sh
eval "$(mcfly init bash)"
eval "$(zoxide init bash)"
```

With these two tool enabled at the same time, when I open a new shell I get this:
`
bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
bash: PROMPT_COMMAND: line 0: `mcfly_prompt_command;;__zoxide_hook'
`
Then, running mcfly:

`
thread 'main' panicked at 'McFly error: Please ensure that MCFLY_HISTORY is set (environment variable not found)', /net/homedirs/arusev/.cargo/registry/src/gitlab.spimageworks.com-dbbe1dfef02d8f03/mcfly-0.5.6/src/settings.rs:261:29
`

If I comment the `zoxide` line in my bashrc then it works. Any ideas which tool cases it?

Thanks.

cantino:
Should fix #159.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -57,7 +57,7 @@ function mcfly_prompt_command {
 }
 
 # Set $PROMPT_COMMAND run mcfly_prompt_command and then any existing $PROMPT_COMMAND.
-PROMPT_COMMAND="mcfly_prompt_command;$(echo $PROMPT_COMMAND | sed 's/^;//')"
+PROMPT_COMMAND="mcfly_prompt_command;${PROMPT_COMMAND#;}"
 
 # If this is an interactive shell, take ownership of ctrl-r.
 if [[ $- =~ .*i.* ]]; then
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -57,7 +57,7 @@ function mcfly_prompt_command {
 }
 
 # Set $PROMPT_COMMAND run mcfly_prompt_command and then any existing $PROMPT_COMMAND.
-PROMPT_COMMAND="mcfly_prompt_command;$PROMPT_COMMAND"
+PROMPT_COMMAND="mcfly_prompt_command;$(echo $PROMPT_COMMAND | sed 's/^;//')"
 
 # If this is an interactive shell, take ownership of ctrl-r.
 if [[ $- =~ .*i.* ]]; then


cantino:
@SafariMonkey, does this seem okay?

SafariMonkey:
Personally, I would double quote the inner `$PROMPT_COMMAND`, because otherwise you're relying on `echo` putting the arguments back together correctly - _probably_ safe, but preferably bypassed altogether.

I tried putting it into [shellcheck.net](https://www.shellcheck.net/):

```bash
#!/bin/bash
PROMPT_COMMAND="mcfly_prompt_command;$(echo $PROMPT_COMMAND | sed 's/^;//')"
```

And got out:

```
$ shellcheck myscript
 
Line 2:
PROMPT_COMMAND="mcfly_prompt_command;$(echo $PROMPT_COMMAND | sed 's/^;//')"
                                       ^-- SC2001: See if you can use ${variable//search/replace} instead.
                                            ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: (apply this, apply all SC2086)
PROMPT_COMMAND="mcfly_prompt_command;$(echo "$PROMPT_COMMAND" | sed 's/^;//')"

$
```

It recommends the quoting of the variable, as I mentioned. It also suggests trying to use the native substitution capabilities built into `bash` instead of using `sed`. Looking at that a little more, I'm pretty sure the following is equivalent to your command:

```bash
PROMPT_COMMAND="mcfly_prompt_command;${PROMPT_COMMAND#;}"
```

Shellcheck is happy with that.

cantino:
Thanks @SafariMonkey! My bash skill level is fairly low, so I really appreciate another set of eyes.

Without the substitution:

```
PROMPT_COMMAND=";echo hi" ./dev.bash
bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
bash: PROMPT_COMMAND: line 0: `mcfly_prompt_command;;echo hi'
```

With the substitution:

```
PROMPT_COMMAND="echo hi" ./dev.bash
hi
bash-3.2$ echo $PROMPT_COMMAND
mcfly_prompt_command;echo hi

```

```
PROMPT_COMMAND=";echo hi" ./dev.bash
hi
bash-3.2$ echo $PROMPT_COMMAND
mcfly_prompt_command;echo hi
```

cantino:
This should fix your issue @alexxbb.

cantino:
For future reference: https://tldp.org/LDP/abs/html/string-manipulation.html

alexxbb:
Thank you @cantino . I will try it as soon as I can.

Issue #50, opened 2019-03-06
Recent commands are hard to find.


ghuls:
A lot of times, a command I have run recently, does not show up when searching for a part of the command.

Also multiline commands does not seem to work.

Pasting this in the terminal and then pasting the following:
```
some_function () {
  printf '1\t2\t3\n' \
    | awk -F '\t' '
    {
        print $1;
        print $2;
        print $3;
    }' \
    | cat 
}
```
In bash without mcfly, pressing the up key, will bring back this full entry (history command):
 2006  some_function () {   printf '1\t2\t3\n'     | awk -F '\t' '
    {
        print $1;
        print $2;
        print $3;
    }'     | cat ; }

With mcfly, this entry is splitted:
```
 1340  some_function () {   printf '1\t2\t3\n'     | awk -F '\t' '
 1341      {
 1342          print $1;
 1343          print $2;
 1344          print $3;
 1345      }'     | cat ; }
```

cantino:
That does sound like a real bug, thank you for reporting it.

cantino:
Are you sure that the normal bash history works correctly with multiline commands? When I try your example without mcfly, it works in the local session, but if I start a new session, it shows up as separate lines in `history`.

ghuls:
Apparently rereading history from bash history does not preserve multiline commands: https://lists.gnu.org/archive/html/bug-bash/2011-02/msg00029.html. It is a pity as I use multiline command quite often.

shelvacu:
I see this is tagged `wontfix`. Would you be willing to merge a pull request for this?

cantino:
Sure, if you can find a solution! I'm not sure how you can fix it. Bash history without timestamps doesn't differentiate single and multiline, or does it?

shelvacu:
Once the history is written with `history -a` then the information is lost, but `history 1` is reliably parseable. You'd need to restructure a bit and pipe it into the mcfly binary for parsing, maybe `HISTTIMEFORMAT="%s " history 1 | mcfly record_history bash` or something.

richvim:
Doesn't presence of a backslash at the end indicate multi-line?

shelvacu:
Multiline commands can also be started from unclosed quotes or parens. I would hope to avoid trying to parse bash commands to see which are part of a multi-line command, as that would be shell & version specific, and hard to get perfect.

cantino:
If you can get it working, I'm very open to a PR.

PR #162, opened 2021-06-20
WIP: Add newline support

shelvacu:
So I've been working on #50 and it sort of feels like things are sprawling out more and more, and I wanted to both contribute what I'd done so far and get feedback before making more changes.

As it stands right now, this PR *does* correctly record commands-with-newline(s) into the database.

The first issue I ran into was that the search interface would not show newlines correctly. I believe I've fixed it, but I haven't tested it rigourously.

While developing, I noticed that the search was running the wrong version of mcfly (installed version in `.cargo/bin` rather than the development version in `target`) despite using `dev.bash`, because the history addendum uses `$MCFLY_PATH` whereas the bind command just calls `mcfly`. I was unable to wrassle the multi-level quoting into working correctly in the bind command with a variable substitution, so I instead removed `$MCFLY_PATH` and relied on `mcfly` calling the right binary. This may not be desireable.

In changing the way the command is read in from history, I cut out some of the code that wrote to a file, filtered, and wrote back out history to then be re-read by `history -cr`. I think for bash the best solution would involve no itermediary files, just the database and `history` calls.

There seems to be some logic for removing history entries starting with `#mcfly`, but I don't see that in my history anyway so I don't understand why it's there. As it is in this PR, that functionality is never used I think.

Let me know your thoughts.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -21,8 +21,7 @@ MCFLY_SESSION_ID="$(dd if=/dev/urandom bs=256 count=1 2> /dev/null | env LC_ALL=
 export MCFLY_SESSION_ID
 
 # Find the binary
-MCFLY_PATH=${MCFLY_PATH:-$(which mcfly)}
-if [ -z "$MCFLY_PATH" ]; then
+if [ -z "$(which mcfly)" ]; then
   echo "Cannot find the mcfly binary, please make sure that mcfly is in your path before sourcing mcfly.bash."
   return 1
 fi
@@ -44,15 +43,11 @@ function mcfly_prompt_command {
     tail -n100 "${HISTFILE}" >| "${MCFLY_HISTORY}"
   fi
 
-  history -a "${MCFLY_HISTORY}" # Append history to $MCFLY_HISTORY.
-  # Run mcfly with the saved code. It will:
-  # * append commands to $HISTFILE, (~/.bash_history by default)
-  #   for backwards compatibility and to load in new terminal sessions;
-  # * find the text of the last command in $MCFLY_HISTORY and save it to the database.
-  $MCFLY_PATH add --exit ${exit_code} --append-to-histfile
-  # Clear the in-memory history and reload it from $MCFLY_HISTORY
-  # (to remove instances of '#mcfly: ' from the local session history).
-  history -cr "${MCFLY_HISTORY}"
+  # Run mcfly with the last history entry on stdin. It will:
+  # * Append the command to $HISTFILE, (~/.bash_history by default)
+  # * Parse out the command and and save it to the database.
+  HISTTIMEFORMAT="%s:" history 1 | mcfly add --exit $exit_code --command-from-stdin
+
   return ${exit_code} # Restore the original exit code by returning it.
 }
 
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -200,6 +200,7 @@ impl<'a> Interface<'a> {
             self.selection = self.matches.len() - 1;
         }
 
+        let mut line_offset = 0;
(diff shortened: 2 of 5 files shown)


shelvacu:
@cantino I'd like your thoughts on this before I go further, not sure if draft pull requests make a notification or not.

I now better understand the `#mcfly:` functionality; In bash >= 4 it is possible to get a full multiline search from `$READLINE_LINE`, but for earlier bash I assume it will just mess up a multi-command when you press C-r

cantino:
Sorry for the delay @shelvacu! You're correct about `#mcfly:` and bash versions <4.

I think it's entirely possible that the current logic to get history into both the shell history and the mcfly database is crufty / unnecessarily complex. It's evolved over time as I worked on this and I've never really revisited it.

mcfly.bash
```diff
@@ -21,8 +21,7 @@ MCFLY_SESSION_ID="$(command dd if=/dev/urandom bs=256 count=1 2> /dev/null | LC_
 export MCFLY_SESSION_ID
 
 # Find the binary
-MCFLY_PATH=${MCFLY_PATH:-$(which mcfly)}
```
cantino, line 24: There was definitely a reason for $MCFLY_PATH to exist, but I can't recall anymore what it was.

mcfly.bash
```diff
@@ -44,15 +43,11 @@ function mcfly_prompt_command {
...
-  #   for backwards compatibility and to load in new terminal sessions;
-  # * find the text of the last command in $MCFLY_HISTORY and save it to the database.
-  $MCFLY_PATH add --exit ${exit_code} --append-to-histfile
-  # Clear the in-memory history and reload it from $MCFLY_HISTORY
-  # (to remove instances of '#mcfly: ' from the local session history).
-  history -cr "${MCFLY_HISTORY}"
+  # Run mcfly with the last history entry on stdin. It will:
+  # * Append the command to $HISTFILE, (~/.bash_history by default)
+  # * Parse out the command and and save it to the database.
+  HISTTIMEFORMAT="%s:" history 1 | mcfly add --exit $exit_code --command-from-stdin
```
cantino, line 49: Does this still work on bash <4?

src/interface.rs
```diff
@@ -232,33 +233,38 @@ impl<'a> Interface<'a> {
...
+            }
 
             if command.last_run.is_some() {
                 write!(
                     screen,
                     "{}",
                     cursor::Goto(
                         width - 9,
-                        (command_line_index as i16 + result_top_index as i16) as u16
+                        (line_offset as i16 + result_top_index as i16) as u16
```
cantino, line 267: I think some logic got lost here around `is_screen_view_bottom`?

mcfly.bash
shelvacu: According to the bash release notes, `HISTTIMEFORMAT` was added in Bash 3, released in 2004. I did a quick test to be sure, it behaves the same as bash 4. Should be good.

src/interface.rs
shelvacu: Indeed, I pushed another commit to fix this.

mcfly.bash
shelvacu: Just leave it out then? I can't find anything that reads it, just things setting it in the dev.* scripts.

cannonpalms:
@shelvacu What's the status of this PR now that @cantino has left a few comments? From your description I can't tell if there is more to be done. You mention that as-is, the PR _does_ correctly handle multi-line commands, but also state that you were sharing the PR as a draft before going further. 

What is left? Perhaps I can help? This bug is a particular nuisance to my workflow as I tend to lean heavily on multi-line commands.

PR #163, opened 2021-06-20
Make ^d delete forward, the readline(5) default.

rbutoi:
It can also send EOF if there's no more input, but this binding is more
useful and there's the others (^z, ^g, ESC) to do the same.

(end of commit message)

Wasn't sure if this was an oversight or intentional, but I find myself instinctively deleting and accidentally closing `mcfly`.


cantino:
Agreed, thanks!

PR #167, opened 2021-06-24
Add GitHub workflows for running clippy and rustfmt

praveenperera:
This will make sure cargo clippy and cargo fmt were run on PR's


cantino:
Thanks @praveenperera!

PR #172, opened 2021-06-28
Make autogenerated release a draft release

praveenperera:

PR #181, opened 2021-08-02
Revert "Make autogenerated release a draft release"

cantino:
Reverts cantino/mcfly#172


cantino:
Reverting at @praveenperera's recommendation.

PR #185, opened 2021-08-13
bash/zsh: prefer unaliased commands

Mic92:
Users might use aliases in their bashrc/zshrc.
`command` calls the command rather the alias.
Also remove unneeded env command.


cantino:
Thank you!

PR #190, opened 2021-08-24
Avoid variable substitutions

### LINKED ISSUE #176: Command line values are incorrectly parsed for substitutions
Unexpected environment variable substitutions occur on escaped history lines when pressing enter after the previous command included an escaped substitution-like string.

image

I mixed up my quotes and ended up with a file like the above. I can move or delete it without any issues, but McFly prints an error which suggests it is trying to perform substitution on the input, despite it being escaped.

This is an obscure issue, but I figured it might highlight other potential issues with the way history lines are handled, and might present a privilege escalation risk.

cantino:
Should fix #176. @CreativeCactus, is it easy for you to confirm?

### DIFF
--- a/src/path_update_helpers.rs
+++ b/src/path_update_helpers.rs
@@ -4,8 +4,7 @@ use std::path::Path;
 use unicode_segmentation::UnicodeSegmentation;
 
 pub fn normalize_path(incoming_path: &str) -> String {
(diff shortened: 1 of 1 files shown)


cantino:
Okay, just released v0.5.9.

Issue #183, opened 2021-08-09
Option to priorize exact matches over fuzzy ones


yangm97:

cantino:
I assume you mean when `MCFLY_FUZZY` is enabled?

yangm97:
yes

cantino:
Seems reasonable.

nedbat:
I like fuzzy matching so that I don't have to remember the exact punctuation in a file name. But I'm finding that using it completely removes the claimed benefits of McFly's intelligence:
<img width="1125" alt="image" src="https://user-images.githubusercontent.com/23789/132209353-4ea0dba3-3bc2-439a-a81e-db2e9325d379.png">
The command I want is the one I ran 10 minutes ago, at position 6.  The top choice isn't even the best fuzzy match for the words I've typed.  Am I doing something wrong? I have these settings:
```
$ env | grep MCFLY
MCFLY_FUZZY=true
MCFLY_RESULTS=30
MCFLY_HISTORY_LIMIT=10000
MCFLY_SESSION_ID=UzHARz6EfjOqhxy8VIvT9Bcg
MCFLY_HISTORY=/var/folders/10/4sn2sk3j2mg5m116f08_367m0000gq/T/mcfly.XXXXXXXX.EbZr6lxT
```

cantino:
I don't personally use fuzzy matching because I agree that it's of lower quality.

nedbat:
Is there some way to improve it? It seems a shame to offer a setting which seems to negate the primary claim of the tool (intelligent history).

cantino:
I'd be open to contributions that improve it. It was contributed by a user and isn't a feature I use myself. I prefer the non-fuzzy matching for how I tend to use mcfly.

dmfay:
With 0.5.10 just out the fuzzy experience should be dramatically improved. If you have MCFLY_FUZZY=true you'll start with a "fuzzy factor" of 2, or you can set the environment variable to another integer value. Higher values of MCFLY_FUZZY favor shorter _and_ earlier matches; 0 turns it off.

In my testing a fuzzy factor of 1 didn't do quite enough to prioritize what I was searching for, and 10+ weighed brevity and start position too heavily over the built-in rank. As I mentioned in the readme I expect the best results to be in the 2-5 range, but I also only have my own history to test with. If you have the time to try a few different settings please report how it works out for you and what MCFLY_FUZZY value you settle on!

alfonz19:
[...]

PR #203, opened 2021-10-28
Make MCFLY_FUZZY a tuneable int

dmfay:
Experimenting based on discussion in #183

No extra weight:
fuzzy-factor-1

No elaboration needed, it's not great.

Shorter match weighted x2:
fuzzy-factor-2

Immediate improvement, with actual ssh commands at the top and most other sorting fairly reasonable. The `./target/relea[se/mcfly search]` match is ahead of `vi [sql.zsh]`, but I've definitely run the former several times in this directory (and the latter never).

Shorter match weighted x10:
fuzzy-factor-10

There's a point of diminishing returns here and 10 may be past it; `./target/relea[se/mcfly search]` is pushed way down, and even higher values (x20, x50) don't look much different.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -65,7 +65,7 @@ pub enum HistoryFormat {
 pub struct Settings {
     pub mode: Mode,
     pub debug: bool,
-    pub fuzzy: bool,
+    pub fuzzy: i16,
     pub session_id: String,
     pub mcfly_history: PathBuf,
     pub output_selection: Option<String>,
@@ -104,7 +104,7 @@ impl Default for Settings {
             refresh_training_cache: false,
             append_to_histfile: false,
             debug: false,
-            fuzzy: false,
+            fuzzy: 0,
             lightmode: false,
             key_scheme: KeyScheme::Emacs,
             history_format: HistoryFormat::Bash,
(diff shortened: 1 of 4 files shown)


dmfay:
I had the idea of including match start position in the weighting, so earlier as well as shorter matches are preferred. That makes a big difference already, but a `FUZZY_FACTOR` of 2 still improves on 1.

src/settings.rs
```diff
@@ -404,8 +404,18 @@ impl Settings {
                     settings.results = results;
                 }
 
-                settings.fuzzy =
-                    search_matches.is_present("fuzzy") || env::var("MCFLY_FUZZY").is_ok();
+                if let Ok(fuzzy) = env::var("MCFLY_FUZZY") {
+                    if let Ok(fuzzy) = i16::from_str(&fuzzy) {
+                        settings.fuzzy = fuzzy;
+                    } else {
+                        settings.fuzzy = 2;
```
cantino, line 411: I think this needs to be something like (in pseudocode) `if fuzzy.to_lowercase() != "false" { settings.fuzzy = 2; }`

dmfay: oops, you're right -- fixed.

cantino:
Thanks!

cantino:
Released in 0.5.10.

PR #214, opened 2021-12-16
Add zinit install method to readme

sykuang:
Thanks for the awesome jobs.
I use zinit as the Zshell plugin manager. With the example below mcfly can be downloaded and installed automictically. Also I can enter the shell with feeling the delay cause running McFly delay.
Hoping this can help other guys.


cantino:
Thanks @sykuang

PR #216, opened 2021-12-20
use `command -v` instead of `which`

Efreak:
which now outputs annoying deprecation notice in bash. In test, fish does the same but zsh does not--however it might? either way, `command v` works in zsh.


cantino:
Thanks @Efreak!

PR #228, opened 2022-01-23
Avoid casting the ulong as a u32

### LINKED ISSUE #223: Strange character results when using Mcfly
M1 Mac: 12.1 (Monterey)
iTerm: 3.4.15
zsh: 5.8
zsh plugins: {zsh} | -z, -autosuggestions, autojump, git, vi-mode
other plugins: tldr

When I run `mcfly search {keyword}` the history bar pops up like normal. However, any attempt to complete something from this menu results in a strange "illegal byte sequence" error. The `zvm_update_repeat_commands` error is the result of `vi-mode` choking on the byte sequences (but removing that plugin does not resolve the issue). I have removed all plugins and witnessed the same behavior.

Please see the two screenshots I have added below.

Screen Shot 2022-01-09 at 10 23 15 AM

After hitting `Enter` or `Tab`

Screen Shot 2022-01-09 at 10 23 37 AM

cantino:
Fix #223 with @arunpersaud's solution.

### DIFF
--- a/src/fake_typer.rs
+++ b/src/fake_typer.rs
@@ -1,12 +1,14 @@
+use libc;
+
 // Should we be using https://docs.rs/libc/0.2.44/libc/fn.ioctl.html instead?
 extern "C" {
-    pub fn ioctl(fd: i8, request: u32, arg: *const u8) -> i8;
+    pub fn ioctl(fd: libc::c_int, request: libc::c_ulong, arg: ...) -> libc::c_int;
 }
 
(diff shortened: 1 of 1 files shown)


cantino:
Oh wait, this must be OS-dependent?

```
  --> src/fake_typer.rs:11:30
   |
11 |         if unsafe { ioctl(0, libc::TIOCSTI, a) } < 0 {
   |                              ^^^^^^^^^^^^^ expected `u32`, found `i32`
```

PR #232, opened 2022-02-03
Iss226 - Copy Highlighted Command to system clipboard

### LINKED ISSUE #226: Feature request: Key-binding to copy highlighted command to system clipboard
Hi, 
Is there a way to add a keybinding to copy a highlighted command from mcfly's heads-up display directly to the system clipboard? Presently, the only way I can achieve this consistently across several terminal emulators (kitty, gnome-terminal etc.) is by using the edit option (pressing TAB) and then copying the result into the emulator's internal buffer (using Ctrl+U). 
Thanks!

michaelnavs:
Closes #226

As discussed in issue 226, I installed the rust-clipboard crate to easily add text to the system clipboard. I updated the menu interface text to display the new keybinding for the user to know which keys to press to copy the selected command. I decided to use 'Ctrl+y' in both emacs and vim keybindings. I could not use the default keybindings from either one, so I decided it would be best to use the same keybinding for both. Also, I decided that the user should be able to use the keybinding in both insert and command mode of vim.

Love to hear your feedback @cantino 🤘🏼

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,6 +73,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
+
+[[package]]
 name = "bstr"
 version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @michaelnavs, this is a cool feature! I just tested it and it worked on my Mac. A few thoughts:

- Do you think we should use a function key for copy instead of ctrl-y? I'm honestly not sure, I'd be okay with either but then it would match delete.
- I think it should provide some visible feedback when you copy, to let you know that it worked.
- Have you tested on Windows / Linux?

michaelnavs:
- I honestly didn't even think about using a function key, but it is a good idea since it would be consistent for all keybindings and match the delete command. Plus, it would clean up the menu interface UI. 
- I thought about the adding some feedback, but I don't know how you would want that to be implemented. If you have any ideas, I would be more than happy to add it.
- I am also on Mac, so I know it works on Mac. I have a windows machine and can test it, but I won't be able to do it until Sunday as I will be out of town until then. I do not have a way to test directly on Linux, but I can install a vm and test it if would like me to.

michaelnavs:
Tested on Linux (Ubuntu), worked as expected. It behaves just like MacOS.
Tested it on Windows with WSL Ubuntu and I run into an error on startup. 
Capture

The error comes from:

```rust
    pub fn new(settings: &'a Settings, history: &'a History) -> Interface<'a> {
        Interface {
            history,
            settings,
            input: CommandInput::from(settings.command.to_owned()),
            selection: 0,
            matches: Vec::new(),
            debug: settings.debug,
            run: false,
            delete_requests: Vec::new(),
            menu_mode: MenuMode::Normal,
            in_vim_insert_mode: true,
            clipboard_context: ClipboardProvider::new().unwrap(), // error here
        }
    }
```

michaelnavs:
It happens on boot

PR #235, opened 2022-02-14
Don't prepend extra copies to prompt command

nfultz:
When starting subshells, extra copies of mcfly_prompt_command can get preprended. 
Instead, check first and only prepend if not found.


cantino:
Good catch, thank you @nfultz!

cantino:
https://github.com/cantino/mcfly/releases/tag/v0.6.0 released!

Issue #239, opened 2022-02-24
Sourcing .zshrc didn't rebind Ctrl+R for me, had to restart kitty


gthb:
New to McFly, I followed the installation instructions in the readme (in Kitty, running zsh), and in this line:

>  Run `. ~/.bashrc` / `. ~/.zshrc` / `source ~/.config/fish/config.fish` or restart your terminal emulator.

... I tried the former, just sourcing `.zshrc`. But Ctrl+R still just invoked the familiar old `bck-i-search` experience.

I tried to troubleshoot that for a while, hunt down where the relevant `bindkey` was invoked etc., but in the end I gave up and restarted Kitty. After that, Ctrl+R invoked McFly as expected.

Is it expected behavior that sourcing `.zshrc` will not suffice in some circumstances? Then maybe change that line to say something like “Run `. ~/.bashrc` [...] to reload your shell configuration, and then try Ctrl+R. If that doesn't bring up McFly, restart your terminal emulator”.

But if sourcing shell config _is_ always supposed to be enough, then maybe there's some initialization problem to hunt down?

cantino:
I wonder if your .zshrc has some code that exits early if it has been sourced before. In general I just restart the shell when installing.

gthb:
Tried to test that by commenting out the `eval` line, starting a new terminal in kitty, verifying that Ctrl-R was not invoking McFly, then uncommenting the `eval` line and surrounding it by echos:
```
echo 'evaling mcfly'
eval "$(mcfly init zsh)"
echo 'done evaling mcfly'
```
and then sourcing .zshrc:

```
$ source ~/.zshrc
evaling mcfly
done evaling mcfly
```

so it did run ... but, uh, then McFly worked. So this failed to reproduce. 😞 I.e. I can't rule out that what you described was the problem _when_ I got the problem.

I wonder if this had to do with whatever shenanigans underlie [PowerLevel10k's “instant prompt”](https://github.com/romkatv/powerlevel10k#instant-prompt) feature (which, yes, I have enabled at the top of my .zshrc). Maybe that causes some of .zshrc not to run (though their description of the feature doesn't sound like it should).

I notice that [they recommend not to source .zshrc at all](https://github.com/romkatv/powerlevel10k#weird-things-happen-after-typing-source-zshrc):

> ### Weird things happen after typing `source ~/.zshrc`
> 
> It's almost always a bad idea to run `source ~/.zshrc`, whether you are using Powerlevel10k or not.
[...]

Issue #219, opened 2022-01-07
Stop working after sourcing ~/.zshrc


andgineer:
MacOS 11.6.2
iTerm 3.4.14
zsh 5.8 (with only git plugin)
at the end of .zshrc I have `eval "$(mcfly init zsh)"`

I also have fig.io installed

It works just after starting iTerm2, in default and newly created tabs

But after `.~/.zshrc` ^r is binded to standard back search

SamShowalter:
I am also having this issue.

cantino:
Hey @andgineer, I'm not sure why it wouldn't be working.

dasebasto:
Something in your .zshrc is probably binding ^r to standard or custom history search.

I had the same issue.
The culprit line in my .zshrc was:
`source $ZSH/oh-my-zsh.sh`

The bug only occurs when you source .zshrc more than once, because `$(mcfly init zsh)` contains this:

```
# Avoid loading this file more than once
if [[ "$__MCFLY_LOADED" == "loaded" ]]; then
  return 0
fi
__MCFLY_LOADED="loaded"
[...]
  bindkey '^R' mcfly-history-widget
```

Quick fix for me was to add
`bindkey '^R' mcfly-history-widget` *before* `eval "$(mcfly init zsh)` in my .zshrc


I believe the current behavior of `eval "$(mcfly init zsh)` should be changed. It should:
- always bind ^R
- not contain return statements that have the nasty side effect of skipping all the following lines in .zshrc

andgineer:
Thank you!
In fact it only help when I placed `bindkey` AFTER `source $ZSH/oh-my-zsh.sh`

Jayphen:
I have the same problem (also with oh-my-zsh). I have tried putting `bindkey` before and after sourcing `oh-my-zsh`, but without luck. ctrl-r always get rebound back to the default search.

tombh:
As mentioned in #254, a quick fix for this is to, instead of `eval "$(mcfly init zsh)"`, just use `source <(mcfly init zsh)`.

PR #255, opened 2022-05-08
Intermediate `source <(mcfly init *sh)` during initialisation. Fixes #254, #219, #239

### LINKED ISSUE #254: Use `source <(mcfly init *sh)` instead of `eval "$(mcfly init *sh)"`
Really enjoying this project, thank you 💕

As mentioned in #219 and #239, there is unexpected behaviour when re-running `.zshrc`. As @dasebasto [pointed out](https://github.com/cantino/mcfly/issues/219#issuecomment-1017339164), this is because of:

```zsh
if [[ "$__MCFLY_LOADED" == "loaded" ]]; then
  return 0
fi
__MCFLY_LOADED="loaded"
```

When `return 0` is `eval`'d it returns from the _whole_ `.zshrc` script, thus failing to run any further commands such as binding `bindkey '^R' mcfly-history-widget`, but more importantly failing to run the rest of `.zshrc` without warning.

Of course I totally agree that `source ~/.zshrc` is not _best_ practice. But it is _a_ practice. As such, it should not be unexpectedly broken without warning.

The quickest fix for this is just to update the README to suggest loading `mcfly` with `source <(mcfly init zsh)`. Another idea is to add an intermediate step, for example this is how [starship](https://github.com/starship/starship) does it:

```sh
# Show contents
% starship init zsh
source <(/usr/bin/starship init zsh --print-full-init)
# Run contents
% eval "$(starship init zsh)"
🚀
```

tombh:
As requested in #254.

I don't know anything about Fish, but I assume this issue didn't affect it?

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -2,14 +2,27 @@ use super::settings::InitMode;
 
 pub struct Init {}
 
+// Intermediating a `source <(...)` step is important so that the actual init
+// scripts don't have too much power in `.bashrc` or `.zshrc` when `eval()`'d. For
+// instance eval'ing `return 0` returns from the _entire_ .bashrc/.zshrc script,
+// not just the mcfly init script.
+//
+// Perhaps just asking users to directly `source <(mcfly init bash --print_full_init)`
+// is better, but we now also need to be backward compatible for those that already
+// have `eval "$(mcfly init bash)"` in their config.
+//
+// See: https://github.com/cantino/mcfly/issues/254
+const BASH_SOURCE_STANZA: &str = "source <(mcfly init bash --print_full_init)";
+const ZSH_SOURCE_STANZA: &str = "source <(mcfly init zsh --print_full_init)";
+
 impl Init {
-    pub fn new(init_mode: &InitMode) -> Self {
+    pub fn new(init_mode: &InitMode, is_print_full_init: bool) -> Self {
         match init_mode {
             InitMode::Bash => {
-                Init::init_bash();
+                Init::init_bash(is_print_full_init);
             }
             InitMode::Zsh => {
-                Init::init_zsh();
+                Init::init_zsh(is_print_full_init);
             }
             InitMode::Fish => {
                 Init::init_fish();
@@ -17,13 +30,21 @@ impl Init {
         }
         Self {}
     }
-    pub fn init_bash() {
-        let script = include_str!("../mcfly.bash");
-        print!("{}", script);
+    pub fn init_bash(is_print_full_init: bool) {
+        if is_print_full_init {
+            let script = include_str!("../mcfly.bash");
+            print!("{}", script);
+        } else {
+            print!("{}", BASH_SOURCE_STANZA);
+        }
     }
-    pub fn init_zsh() {
-        let script = include_str!("../mcfly.zsh");
-        print!("{}", script);
+    pub fn init_zsh(is_print_full_init: bool) {
+        if is_print_full_init {
+            let script = include_str!("../mcfly.zsh");
+            print!("{}", script);
+        } else {
+            print!("{}", ZSH_SOURCE_STANZA);
+        }
     }
     pub fn init_fish() {
         let script = include_str!("../mcfly.fish");
--- a/src/main.rs
+++ b/src/main.rs
@@ -100,7 +100,7 @@ fn handle_move(settings: &Settings) {
 }
 
 fn handle_init(settings: &Settings) {
-    Init::new(&settings.init_mode);
+    Init::new(&settings.init_mode, settings.print_full_init);
 }
 
 fn main() {
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -87,6 +87,7 @@ pub struct Settings {
     pub interface_view: InterfaceView,
     pub result_sort: ResultSort,
     pub disable_menu: bool,
+    pub print_full_init: bool,
 }
 
 impl Default for Settings {
@@ -116,6 +117,7 @@ impl Default for Settings {
             interface_view: InterfaceView::Top,
             result_sort: ResultSort::Rank,
             disable_menu: false,
+            print_full_init: false,
         }
     }
 }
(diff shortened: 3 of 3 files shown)


cantino:
Thanks @tombh! Merged. Are you able to check if this has fixed the bug for you?

cantino:
Actually, I'm getting panics when I run `bash` now and I wasn't before.

```
~ RUST_BACKTRACE=1 bash
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', library/std/src/io/stdio.rs:1187:9
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: std::io::stdio::_print
   3: mcfly::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

This is launching bash from zsh, my default shell.

tombh:
I'm primarily a zsh user too, so I may have overlooked something for bash. I just assumed they were similar enough that the both needed the same solution. But I might be wrong. Anyway, I went as far as testing like this from with a bash shell:
```bash
[tombh@tombox mcfly]$ source <(cargo run -- init bash --print_full_init)
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/mcfly init bash --print_full_init`
[tombh@tombox mcfly]$
```

And `CTRL+r` worked fine for me. How are you testing?

cantino:
I built and installed it locally: `cargo build --release && cp ./target/release/mcfly ~/bin/`
My .zshrc includes `eval "$(mcfly init zsh)"` and my .bashrc has `eval "$(mcfly init bash)"`.
And then from zsh, I ran `bash`.

tombh:
I just repeated that exactly, and can't reproduce the bug. Can you paste the contents of your `.bashrc`? And do you think there's anything unusual in your env?

cantino:
That's very odd. I tried adding a newline to your outputs too. Are you on a
Mac? Maybe related to this? https://github.com/bennofs/nix-index/issues/69

Backing up a second, what was the reasoning around not just changing the
init scripts to be wrapped in a conditional if `if ! [[ "$__MCFLY_LOADED"
== "loaded" ]]; then`?

tombh:
I'm on Arch (by the way haha). What are the newlines for? To replicate the difference between `println!` and `writeln!` as mentioned in that nix-index bug?

> what was the reasoning around not just changing the init scripts to be wrapped in a conditional if `if ! [[ "$__MCFLY_LOADED" == "loaded" ]]; then`?

That's certainly one approach. But the advantage of using process substitution, namely `source <(...)`, is that you never have to worry ever again about what gets put in the init scripts, because there's no possible way they can contaminate the user's own `.zshrc` or `.bashrc`. That just seems like the sanest, safest and most respectful approach.

tombh:
Oh, are you on Mac? Because there is indeed some issues mentioned with this kind of init in the Starship code, have a look: https://github.com/starship/starship/blob/17453929091b1d9463f33e0b291436c9213c54d2/src/init/mod.rs#L104

tombh:
If Starship can do it, so can we! Look, this is what they use for lower Bash versions:
  `source /dev/stdin <<<"$(mcfly init bash --print-full-init)"`

Does it help you? Are you Mac?

cantino:
I am on a Mac, yes. `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)`

No, that isn't working for me. Running `source /dev/stdin <<<"$(mcfly init bash --print_full_init)"` returns no errors, but also doesn't register mcfly with the shell.

I'm leaning toward just conditionalizing the scripts for now.

tombh:
There's an interesting comment at the top of that Starship file:

```rust
/*
We use a two-phase init here: the first phase gives a simple command to the
shell. This command evaluates a more complicated script using `source` and
process substitution.

Directly using `eval` on a shell script causes it to be evaluated in
a single line, which sucks because things like comments will comment out the
rest of the script, and you have to spam semicolons everywhere. By using
source and process substitutions, we make it possible to comment and debug
the init scripts.

In the future, this may be changed to just directly evaluating the initscript
using whatever mechanism is available in the host shell--this two-phase solution
has been developed as a compatibility measure with `eval $(starship init X)`
*/
```

So wrapping the init scripts in a big conditional won't solve the bigger issue. Directly `eval()`'ing is just asking for untold trouble, whether now or later when you or any future contributors are unaware of the various restrictions they need to ensure.

I think we're close. Starship is Rust, and they've already successfully solved all these pieces. We've found the cause of your bug, and so just need to get over that final hurdle. If you don't have time to figure it out, then I think the better compromise is to only fallback for Bash <v4.1 users, something like this:

```bash
local major="${{BASH_VERSINFO[0]}}"
local minor="${{BASH_VERSINFO[1]}}"
if ((major > 4)) || {{ ((major == 4)) && ((minor >= 1)); }}; then
	source <(mcfly init bash --print_full_init)
else
	mcfly init bash --print_full_init
fi
```

cantino:
Starship's `${{BASH_VERSINFO[0]}}` stuff isn't working for me, unfortunately. I just created https://github.com/cantino/mcfly/pull/259 as a temporary solution. Mind giving it a look? I unfortunately don't have much time to put into mcfly anymore. I should post looking for co-maintainers.

tombh:
I think you misunderstood my advice in that previous comment. The `${BASH_VERSINFO[0]}` isn't to fix your issue, it's to disable the fix in this PR _just_ for your version of Bash. That way everybody else gets the safer `source <(mcfly init)` behaviour. Bash v4.1 is 12 years old after all 🫤

Also, I didn't realise that Rust is escaping brackets with another bracket `{{`, so that wouldn't have worked as is anyway. I've pushed a commit that actually implements it correctly. And there's no reason why you can't have both this PR and your #259 approach. At least yours will help whilst the `${BASH_VERSINFO[0]}` conditional is serving the bad behaved version.

tombh:
Oh, because this PR is closed the commit isn't automatically showing up. It's here: https://github.com/tombh/mcfly/commit/54984f8bd877dfe10f5766437c73a5bc2aca1595 I'll have to make another PR for it, if you want it.

cantino:
Right, but the `${BASH_VERSINFO[0]}` stuff simply isn't working when I try it.

tombh:
Oh I made a typo, it should be `${BASH_VERSIONFO[0]}` not `${BASH_VERSINFO[0]}`

cantino:
If you want to remake the PR with the correct code, I'm happy to give it another try locally.

PR #256, opened 2022-05-08
Include more linux builds on releases

### LINKED ISSUE #253: Install script attempts to download aarch64-unknown-linux-musl when run on 64 Raspberry Pi
Hi, I have mcfly working great on a number of Linux machines.

I'm trying to install it on a Raspberry Pi running a 64bit OS.  I see from previous posts people have installed mcfly on Pis successfully, using the Armv7 version.  When I try this (or any Arm version), I get a "file not found" error.  I believe I have the path set correctly and the file is being found, but that it is not executable on 64bit Raspberry Pi OS.

I have also tried running the standard install script.  This fails when it attempts to download:  

aarch64-unknown-linux-musl

So it appears an aarch64 version is required for 64bit Pi OS?

Is an aarch64 version available?

Thanks

praveenperera:
Closes #253

Binary list here: https://github.com/praveenperera/mcfly/releases

### DIFF
--- a/.github/workflows/mean_bean_ci.yml
+++ b/.github/workflows/mean_bean_ci.yml
@@ -89,3 +89,6 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - i686-unknown-linux-musl
           - x86_64-unknown-linux-musl
+          - aarch64-unknown-linux-gnu
+          - aarch64-unknown-linux-musl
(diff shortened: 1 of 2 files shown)


cantino:
Thank you!

cantino:
Released in 0.6.1!

PR #257, opened 2022-05-09
Revert "Intermediate `source <(mcfly init *sh)` during initialisation. Fixes #254, #219, #239"

### LINKED ISSUE #254: Use `source <(mcfly init *sh)` instead of `eval "$(mcfly init *sh)"`

cantino:
Reverts cantino/mcfly#255

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -2,27 +2,14 @@ use super::settings::InitMode;
 
 pub struct Init {}
 
-// Intermediating a `source <(...)` step is important so that the actual init
-// scripts don't have too much power in `.bashrc` or `.zshrc` when `eval()`'d. For
-// instance eval'ing `return 0` returns from the _entire_ .bashrc/.zshrc script,
-// not just the mcfly init script.
-//
-// Perhaps just asking users to directly `source <(mcfly init bash --print_full_init)`
-// is better, but we now also need to be backward compatible for those that already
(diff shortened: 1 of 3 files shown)


cantino:
Reverting until we figure out why I'm getting crashes when launching bash.

PR #258, opened 2022-05-09
Get lates release tag name without needing to install `jq`

praveenperera:
We can remove the dependency on `jq`


cantino:
Released in 0.6.1!

PR #259, opened 2022-05-15
Wrap scripts in conditional

cantino:
Short term fix for https://github.com/cantino/mcfly/issues/254?


cantino:
Released in 0.6.1!

PR #262, opened 2022-05-19
Fix vim append command

fabiogibson:
This is a slight vim mode improvement to make insert and append commands work like in vim.

i => Inserts at the current position
I => inserts at the beginning of the line.
a => inserts at one position to the right.
A => inserts at the end of the line.


cantino:
Thanks @fabiogibson!

cantino:
Released in 0.6.1!

PR #263, opened 2022-05-25
implemented feature to switch result sort by key binding

### LINKED ISSUE #238: Feature Request: Switch sort mode LAST RUN and RANK by using menu's interface
It's more convenient to use hot key to switch to different mode if current history output doesn't provide suitable results

navazjm:
Closes #238 
- Added result_sort as an attribute to Interface so we can mutate it. Similar to interface.input
- When creating new instance of Interface, result_sort is set to settings.result_sort.to_owned()
- Added a new function switch_result_sort, which switches between the two result sorts. It is called by pressing F1, regardless of the key scheme.
- Added the clone trait to ResultSort Enum
- Passed in the new interface.result_sort into interface.history.find_matches 
- Finally, I rewrote the MenuMode.text function. I rewrote it because it would have become a mess trying to add an additional match statement to each of the existing match statements. For clarification, if I did not rewrite it, It would look like the following:

```rust
impl MenuMode {
    fn text(&self, interface: &Interface) -> &str {
        match *self {
            MenuMode::Normal => match interface.settings.key_scheme {
                KeyScheme::Emacs => match interface.result_sort {
                    ResultSort::Rank => "McFly | ESC - Exit | ⏎ - Run | TAB - Edit | F1 - Sort (Time) | F2 - Delete",
                    ResultSort::LastRun => "McFly | ESC - Exit | ⏎ - Run | TAB - Edit | F1 - Sort (Rank) | F2 - Delete",
                }
                KeyScheme::Vim => {
                    if interface.in_vim_insert_mode {
                        match interface.result_sort {
                            ResultSort::Rank => "McFly (Ins) | ESC - Cmd | ⏎ - Run | TAB - Edit | F1 - Sort (Time) | F2 - Delete",
                            ResultSort::LastRun => "McFly (Ins) | ESC - Cmd | ⏎ - Run | TAB - Edit | F1 - Sort (Rank) | F2 - Delete",
                        }
                    } else {
                        match interface.result_sort {
                            ResultSort::Rank => "McFly (Cmd) | ESC - Exit | ⏎ - Run | TAB - Edit | F1 - Sort (Time) | F2 - Delete",
                            ResultSort::LastRun => "McFly (Cmd) | ESC - Exit| ⏎ - Run | TAB - Edit | F1 - Sort (Rank) | F2 - Delete",
                        }
                    }
                }
            },
            MenuMode::ConfirmDelete => "Delete selected command from the history? (Y/N)",
        }
    }
```

Also, I think by rewriting this function, reduces the number of places we have to add new key bindings. Above, we would have to change all 6 of the different strings. While with the new rewrite, the programmer has to figure out where to append the new keybinding.

Let me know your thoughts @cantino

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -519,7 +519,10 @@ impl<'a> Interface<'a> {
                 self.input.insert(c);
                 self.refresh_matches();
             }
-            Key::F(1) => self.switch_result_sort(),
+            Key::F(1) => {
+                self.switch_result_sort();
+                self.refresh_matches();
+            },
             Key::F(2) => {
                 if !self.matches.is_empty() {
                     if self.settings.delete_without_confirm {
@@ -576,7 +579,10 @@ impl<'a> Interface<'a> {
                     self.input.insert(c);
                     self.refresh_matches();
                 }
-                Key::F(1) => self.switch_result_sort(),
+                Key::F(1) => {
+                    self.switch_result_sort();
+                    self.refresh_matches();
+                },
                 Key::F(2) => {
                     if !self.matches.is_empty() {
                         if self.settings.delete_without_confirm {
@@ -645,7 +651,10 @@ impl<'a> Interface<'a> {
                 Key::Char(_c) => {
 
                 }
-                Key::F(1) => self.switch_result_sort(),
+                Key::F(1) => {
+                    self.switch_result_sort();
+                    self.refresh_matches();
+                },
                 Key::F(2) => {
                     if !self.matches.is_empty() {
                         if self.settings.delete_without_confirm {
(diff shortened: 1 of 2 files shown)


cantino:
Hey @tehmj, thanks for the contribution! When I tried it, pressing F1 updates the menu, but it doesn't actually resort the list until I start typing. Do you think it should immediately trigger a resort?

navazjm:
I think it should trigger a resort. I mentioned this in the original issue, but did not receive an answer and forgot to include it in my PR. I will try to make the changes and get them committed. @cantino

cantino:
Sorry for the delay. Thanks for this nice addition @navazjm!

cantino:
Released in 0.6.1!

PR #272, opened 2022-07-30
Fix the handling of open-quote strings in fish

hivehand:
In fish, with McFly initiated, hitting Control-R when the existing
command line contains an unmatched quote currently triggers a fish
error, e.g.:

```
❯ echo "Hi- (line 1): Unexpected end of string, quotes are not balanced
/opt/homebrew/bin/mcfly --mcfly_history /dev/null --history_format fish search -o $mcfly_output -- echo "Hi
                                                                                                        ^
```

Fortunately, the solution is simple: use fish's own `string escape` to
escape the output of `commandline` before passing it to `mcfly` as an
argument.


cantino:
Thanks @hivehand!

Issue #85, opened 2020-05-24
Inline suggestions in zsh like zsh-autosuggestions


cantino:
What would it take to make a zsh-autosuggestions completion hook for McFly?

cantino:
Also might be able to improve bash bindings using something like https://stackoverflow.com/questions/57811962/bind-up-arrow-key-to-fzf-command

NightMachinary:
@cantino You might even be able to reuse zsh-autosuggestions itself. They added the ability to add “providers” to the suggestions (like inline suggestions for completions), and I think you can use that abstraction to easily add your history suggestions. (I have no further knowledge, I guess you can just open an issue and ask them.)

Mic92:
Does not look very complex: https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/strategies/history.zsh#L31

Could mcfly search just output all suggestions as lines instead of opening a TUI when it detects a pipe on stdout?

Mic92:
The hacky work around is using this:

```
╭─[~/.homesick/repos/dotfiles/zsh/zsh-autosuggestions]─[tags/v0.6.4]
╰─ % mcfly search --output-selection /tmp/foo1 foo </dev/null

╭─[~/.homesick/repos/dotfiles/zsh/zsh-autosuggestions]─[tags/v0.6.4]
╰─ % cat /tmp/foo1
mode display
commandline foo
```

however it still tries to spawn a menu for a split-second.

praveenperera:
@cantino do you mean like what this does? https://github.com/Aloxaf/fzf-tab

Mic92:
Not quite, Auto-suggestions show up without tabs.

This hacky line extracts suggestions:
```
$ suggestion=$(script -q -O /dev/null -c "./result/bin/mcfly search -o /proc/self/fd/3 test </dev/null" 3>&1 >/dev/null)
$ echo $suggestion
mode display
commandline test
```

PR #276, opened 2022-08-16
`no_ui` option that directly writes the first suggestion to a file

chengyuhui:
This should make #85 eaiser to do, since we do not need to deal with the UI popping up. Bypassing the UI should also make `mcfly` start a little faster.

I was originally wanting to add support for `-` in `output_selection`, but since we are doing `print!` everywhere, this will not be effective as standard output may contain all kinds of messages.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,6 +50,41 @@ fn handle_addition(settings: &Settings) {
 
 fn handle_search(settings: &Settings) {
     let history = History::load(settings.history_format);
+
+    if settings.no_ui {
+        history.build_cache_table(
+            &settings.dir.to_owned(),
+            &Some(settings.session_id.to_owned()),
+            None,
+            None,
+            None,
+            settings.limit.to_owned(),
+        );
+        let matches = history.find_matches(
+            &settings.command,
+            settings.results as i16,
+            settings.fuzzy,
+            &mcfly::settings::ResultSort::Rank,
+        );
+
+        let cmd = match matches.first() {
+            Some(cmd) => cmd.cmd.to_owned(),
+            None => settings.command.to_owned(),
+        };
+        let cmd_t = cmd.trim();
+        if cmd_t.is_empty() {
+            return;
+        }
+
+        let out = format!("mode display\ncommandline {}\n", cmd_t);
+        if let Some(path) = &settings.output_selection {
+            fs::write(path, out)
+                .unwrap_or_else(|err| panic!("McFly error: unable to write to {}: {}", path, err));
+        }
+        
+        return;
+    }
+
     let result = Interface::new(settings, &history).display();
     if let Some(cmd) = result.selection {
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @chengyuhui. Can you explain a bit more what you hope to use this for?

src/main.rs
```diff
@@ -50,6 +50,41 @@ fn handle_addition(settings: &Settings) {
...
+        let matches = history.find_matches(
+            &settings.command,
+            settings.results as i16,
+            settings.fuzzy,
+            &mcfly::settings::ResultSort::Rank,
+        );
+
+        let cmd = match matches.first() {
+            Some(cmd) => cmd.cmd.to_owned(),
+            None => settings.command.to_owned(),
```
cantino, line 72: It seems like if it cannot find a match, it should return nothing, not the search query?

src/main.rs
```diff
@@ -50,6 +50,41 @@ fn handle_addition(settings: &Settings) {
...
+        };
+        let cmd_t = cmd.trim();
+        if cmd_t.is_empty() {
+            return;
+        }
+
+        let out = format!("mode display\ncommandline {}\n", cmd_t);
+        if let Some(path) = &settings.output_selection {
```
cantino, line 80: Instead of requiring `--output_selection`, you could allow two options:

1. With `--output_selection`, write to the `--output_selection` selection file as you're doing now.
2. Without output_selection, instead just print one or more matches to STDOUT for scripts to parse.

src/main.rs
chengyuhui: Yes, I somehow misunderstood what the original UI code was doing.

src/main.rs
chengyuhui: It was originally written this way, however I found that we are happily `print!`-ing many messages to STDOUT, which (given the current output format) might mess up parsing for simple scripts.

cantino: We should only be using print! to output when there's an error. Those should probably be converted to use STDERR.

src/main.rs
cantino: This is still returning the search query, right?

src/main.rs
chengyuhui: We should update all `print!` to `eprint(ln)!` before merging this.

cantino: Agreed!

cantino:
Could to proposal in #307, if it allowed a `--limit` on the number of returned matches, also meet this PR's goals?

chengyuhui:
Is it suitable for use when it is called on every key entered into the terminal? If it is, and it does not show the UI, then it should work well.

cantino:
Are you still working on this?

plawley:
just a bump that there is definitely interest in this no-ui direct suggestion capability.  
-zsh-autosuggestions integration is a bit of a killer app for this type of tool.

PR #280, opened 2022-09-04
Use crossterm instead of termion and add powershell support

jtschuster:
This builds off of the PR @dmfay made a while back to use crossterm, adds workarounds for windows related to paths and the fake_typer, and a powershell init script. Behavior should be identical as far as I can tell, but I may have missed something.


cantino:
Thank you for working on this @jtschuster!

Is there anyone who uses Powershell who could test this?

jtschuster:
FWIW I've been using this personally for the past few months and haven't really hit issues (though I didn't have any baseline usage on Linux or Mac with the original version)

cantino:
Is this ready?

jtschuster:
Not quite, there are still some scenarios with multiline commands that don't work yet.

jtschuster:
@cantino this is finally ready

cantino:
Released, do you mind testing it? You may want to build master and make sure my change of 

```diff
 [target.'cfg(windows)'.dependencies]
-autopilot = {git="https://github.com/autopilot-rs/autopilot-rs"}
+autopilot = "0.4.0"
```

works okay.

cantino:
cargo publish wouldn't work with a git dependency, it only supports
versioned dependencies unfortunately.

jtschuster:
> cargo publish wouldn't work with a git dependency, it only supports versioned dependencies unfortunately.

Ah, okay. It does work with the powershell init script since that uses a file output by default, but it does require the file. We could just remove the dependency and require -o on windows, or copy the necessary source from autopilot. I think it would be a fairly thin slice to get just windows typing capability.

jtschuster:
I don't see any windows builds on the releases page, should I be looking somewhere else? My local build of main is working on my machine, though.

cantino:
@praveenperera do you know how to enable windows builds?

cantino:
Could we add usage instructions to the README @jtschuster? https://github.com/cantino/mcfly/issues/400

PR #291, opened 2022-10-13
Fix #288 - enable compatibility with `mktemp` from `uutils/coreutils`

### LINKED ISSUE #288: mktemp (from coreutils) using relative directory instead of temp directory
I'm using `mktemp` (0.0.15) from https://github.com/uutils/coreutils

Using `mktemp -t mcfly.XXXXXXXX` creates the file in the local directory instead of the temp directory.

If I use `mktemp -p mcfly.XXXXXXXX` it properly places the files in `$TMPDIR`.

This is likely an edge case. I'm not sure if you want to try to fix and if so, how to.

jhult:
Per #288, use:

`mktemp ${TMPDIR:/tmp}/mcfly.XXXXXXXX`

instead of:

`mktemp -t mcfly.XXXXXXXX`

This enables compatibility with `mktemp` from [uutils/coreutils](https://github.com/uutils/coreutils).

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -35,7 +35,7 @@ if [[ -t 0 ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
 
     # Populate McFly's temporary, per-session history file from recent commands in the shell's primary HISTFILE.
     if [[ ! -f "${MCFLY_HISTORY}" ]]; then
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @jhult!

cantino:
I had to fix this in https://github.com/cantino/mcfly/commit/78a5422e7a2bd4213b13691183c020361aea608f because I told you the wrong syntax for defaulting a variable.

PR #297, opened 2022-11-08
added option to reset cursor selection when refreshing matches

### LINKED ISSUE #290: Issue : Deleting an entry makes the cursor go back to the top
When cleaning-up history, it's frequent that we need to delete multiple contiguous lines/entries.

This could be super quick with repeated strokes on F2 (if `delete_without_confirm` is enabled).
Unfortunately, it's not convenient, because, after deletion, the cursor goes back to the top item.

Suggestion: after deletion of entry on line N, line N (and the corresponding entry) should stay selected (unless N was the last item, in which case line N-1 should be selected).
This is basically the standard behavior of all UX where user can select and delete items in a list using up/down arrows and Del.

By the way, is there a way to always enable  `delete_without_confirm` on Ctrl-R ?

navazjm:
Closes #290

Add a boolean argument to refresh_matches

if true, we reset the cursor selection to the top of the list of commands
if false, the cursor selections stays in the current position in the list of commands

Only deleting a command calls refresh_matches(false). All other commands call refresh_matches(true).

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -386,12 +386,14 @@ impl<'a> Interface<'a> {
                 self.delete_requests.push(command.cmd.clone());
             }
             self.build_cache_table();
-            self.refresh_matches();
+            self.refresh_matches(false);
         }
     }
 
-    fn refresh_matches(&mut self) {
-        self.selection = 0;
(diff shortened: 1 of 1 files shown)


cantino:
Thanks @navazjm!

PR #304, opened 2022-11-28
use realpath

### LINKED ISSUE #292: command -v when in root directory
I found a bug introduced in #216. It turns out when you are in the root directory (`/`) and you have `mcfly` installed in `/bin` (on a lot of Linux distros that is symlinked to `/usr/bin`) then `command -v mcfly` returns `./bin/mcfly`. There is no option to `command` to make it return an absolute path. However, wrapping the call in `realpath` fixes the issue. 

The original change cites a deprecation warning in `which`. I have been unable to reduce this warning (using Arch Linux). Unfortunately due to the name of the command being such a common english word it's very difficult to search for why this warning appears. Reading the man page says nothing about a deprecation. Maybe the OP of #216 was using Mac and there is something in BSD that is causing it?

Either way, I think it should either use `realpath` around `command -v` or go back to using `which` (maybe without using fish's `command` and relying on `$PATH`), as it's a ubiquitous command across Mac & Linux. The existing behavior is very bad.

EDIT: I just noticed that `mcfly` uses `command -v` in all of its shell scripts, and it seems that, at least in the case of bash, it suffers from the same issue mentioned above. On Linux it takes a `-p` argument that prints the absolute path, but that's a POSIX standard command so I don't know what the support is on Mac.

Is there a problem with just using plain `which` and relying on it being in `$PATH`?

cantino:
Attempting to fix #292

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -17,7 +17,7 @@ if [[ -t 0 ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
   export MCFLY_SESSION_ID
 
   # Find the binary
-  MCFLY_PATH=${MCFLY_PATH:-$(command -v mcfly)}
+  MCFLY_PATH=${MCFLY_PATH:-$(realpath "$(command -v mcfly)")}
   if [ -z "$MCFLY_PATH" ]; then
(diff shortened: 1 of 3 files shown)


cantino:
Should be fixed by #309 instead.

Issue #158, opened 2021-06-15
Fzf support?


NightMachinary:
I like fzf’s search syntax. Is it possible to use fzf for the frontend?

cantino:
Hey @NightMachinary! Not currently, no.

thirteen37:
Would be interesting to decouple the various components: the backend tracking of commands and generation of possible completion candidates, vs the actual frontend TUI widget. I would love a ton more config options for the UI, then I realized I'm basically asking for fzf.

lilianmoraru:
For those stumbling on this issue, check this comment: https://github.com/cantino/mcfly/pull/307#issuecomment-1368072336
Or directly the project: https://github.com/bnprks/mcfly-fzf

@NightMachinery @thirteen37 @sagarkar10

PR #307, opened 2022-12-07
Add support for fzf as a frontend UI

bnprks:
This pull request adds an option to use [fzf](https://github.com/junegunn/fzf) as the front-end UI, addressing issue #158 

If the environment variable `MCFLY_FZF` is defined at init time, then the fzf UI will be bound to Ctrl+R rather than the mcfly interface

This adds fzf support by:
1. Make a command `mcfly fzf` which just prints the command history as text to stdout
2. Modify the binding code in the shell init scripts so that if `MCFLY_FZF` is defined then Ctrl+R will instead run a copy of fzf reading its input from the `mcfly fzf` command with reasonable UI defaults

The only limitations of note are:
1. The MCFLY_FZF variable is only evaluated at init time. Due to the way that bash Ctrl+R is handled, it's not clear how to make the Ctrl+R binding change if the user sets MCFLY_FZF after init. 
     - It looks doable to have hot-reload of MCFLY_FZF in zsh and fish, but this binding in mcfly.bash seems hard to make support hot-reloading: https://github.com/cantino/mcfly/blob/master/mcfly.bash#L70-L76
2. Deletion is not supported within the fzf interface, though running `mcfly search` manually should work fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,25 @@ fish:
 ```bash
 set -gx MCFLY_RESULTS_SORT LAST_RUN
 ```
+### Integration with fzf
+To use `fzf` as the search UI for `Ctrl+R` in terminal, set the environment variable
+`MCFLY_FZF` prior to loading mcfly via `mcfly init`. `fzf` must be installed separately 
+([instructions](https://github.com/junegunn/fzf#installation)).
+
+bash / zsh:
+```bash
+export MCFLY_FZF=TRUE
+```
+
+fish:
+```bash
+set -gx MCFLY_FZF TRUE
+```
+
+Tips:
+- The `fzf` interface does not support deleting history entries. Manually run `mcfly search` to access the standard UI.
+- Modify the environment variable `FZF_CTRL_R_OPTS` to pass custom flags for the
+  `fzf` interface
 
 ### Database Location
 
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -63,22 +63,50 @@ if [[ -t 0 ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
 
   # If this is an interactive shell, take ownership of ctrl-r.
   if [[ $- =~ .*i.* ]]; then
-    if [[ ${BASH_VERSINFO[0]} -ge 4 ]]; then
-      # shellcheck disable=SC2016
-      bind -x '"\C-r": "echo \#mcfly: ${READLINE_LINE[@]} >> $MCFLY_HISTORY ; READLINE_LINE= ; mcfly search"'
-    else
-      # The logic here is:
-      #   1. Jump to the beginning of the edit buffer, add 'mcfly: ', and comment out the current line. We comment out the line
-      #      to ensure that all possible special characters, including backticks, are ignored. This commented out line will
-      #      end up as the most recent entry in the $MCFLY_HISTORY file.
-      #   2. Type "mcfly search" and then run the command. McFly will pull the last line from the $MCFLY_HISTORY file,
-      #      which should be the commented-out search from step #1. It will then remove that line from the history file and
-      #      render the search UI pre-filled with it.
-      if set -o | grep "vi " | grep -q on; then
-        bind "'\C-r': '\e0i#mcfly: \e\C-m mcfly search\C-m'"
+
+    # Function to perform ctrl-r binding for fzf. Adapted from junegunn/fzf shell/key-bindings.bash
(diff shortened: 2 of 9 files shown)


bnprks:
One additional limitation I have since thought of:

3. The statistics for selected autocompletes will not be updated

I think this is doable, but will require making the functionality of the History `record_selected_from_ui` function callable from a command-line invocation. 

In general, I'd be happy to take input on whether this change is desirable for mcfly, and if so how to make it minimally invasive on the rest of the code base.

cantino:
Hey @bnprks, I'm trying to decide if this feels like a good direction for mcfly. It feels like it adds unneeded complexity.

Does `mcfly fzf` print the entire command history— e.g., it just dumps the sqlite db? Would a better plan be to have a fzf shell wrapper that talks to sqlite? Install mcfly for the command recording, but override it's ctrl-r binding to use your script and fzf?

bnprks:
Yes, `mcfly fzf` prints the whole command history (up to `MCFLY_HISTORY_LIMIT`). I've been using this on 10k+ command histories without trouble. I believe `contextual_commands` temporary table creation also requires a scan through the same amount of history on each `mcfly search` invocation.

Of course an external command would be able to similarly read the mcfly database without modifying the core mcfly code. In this case, I think two aspects mean the wrapper code will have to contain a large fraction of mcfly logic internally:

1. Providing the option to sort by mcfly's neural network ranking requires the code to evaluate neural network rankings
2. Marking the selected commands also requires access to some other database insertion logic

It would probably be feasible to implement this using mcfly as a dependency in order to pull in this mcfly logic

bnprks:
Yeah, I think you're understanding the goals right. Thinking about it more, it might be possible to split entirely into a new repo. I'll test it out before completely abandoning this pull request, but I think since mcfly is on crates.io it might work to have a `mcfly-fzf` binary that implements the `fzf` commands, with mcfly as a dependency. Then that separate repo could include both the keybindings and the database functionality and mostly avoid duplicating mcfly code.

I think the main question there would be whether mcfly's public API exposes all the needed functionality, but I think it does

danihodovic:
@cantino is it currently possible to pipe results between mcfly and fzf? Use mcfly as the search "backend" while fzf handles the UI.

cantino:
@danihodovic, It isn't possible today, but with the new options to `search` that we're discussing, you could have a script run `mcfly search "foo bar"` and then pipe that output into fzf.

bnprks:
My first tests on separating this into a new binary seem promising -- basically all rust functions called in `main.rs` can be accessed publicly by a 3rd-party tool. The anticipated usage would be users run `eval $(mcfly init bash)`, then run `eval $(mcfly-fzf init bash)`, and there's a separate `mcfly-fzf` binary that contains the relevant functionality to read + dump the database.

The only change that would be needed in the main `mcfly` repo for this approach is updating the schema versioning logic to throw an error if it encounters a schema that's too new relative to the mcfly version. This would prevent conflicts in case a user's `mcfly` binary version diverged substantially from their `mcfly-fzf` version.

cantino:
@bnprks Do you think it would be best to add the mcfly search command to dump text so that you're not depending on an internal API?

lilianmoraru:
Please take https://github.com/lotabout/skim into consideration as well, so it's not completely tied into `fzf`.
Maybe it helps you decide on a better method of wrapping these and other/future tools.

cantino:
I assume skim would be able to handle dumping text from `mcfly search foo --raw` or similar as well? In which case, that supports my desire to do it that way.

bnprks:
I've had success making a companion tool to mcfly here: https://github.com/bnprks/mcfly-fzf

I think compatibility with mcfly should be fairly straightforward to maintain even if the mcfly rust API changes. The only point of direct interaction between the `mcfly` binary and the `mcfly-fzf` binary is through the sqlite3 database. Given that the mcfly db schema has not changed in 4 years I think keeping `mcfly-fzf` in sync with future schema changes should not be too challenging.

I think it is most logical to have a companion tool specialized to providing the fzf interface. Being able to specialize the output and functionality to fzf's CLI options has been quite valuable and I'm not sure a generic text-dump option in mcfly would be able to provide as good a user experience without becoming tied to assumptions about how fzf works specifically.

Of course I'd be happy to help port some changes into the main mcfly binary if desired, but the companion-tool approach might scale better if people want to add support for other fuzzy search CLIs in the future.

cantino:
Nice work @bnprks!

PR #308, opened 2022-12-08
Avoid exporting HISTFILE

cantino:
Hopefully fixes 249.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -5,10 +5,10 @@
 if [[ -t 0 ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
   __MCFLY_LOADED="loaded"
 
-  # Ensure HISTFILE exists.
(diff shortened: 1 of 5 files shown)


cantino:
@dithpri (or @AndrewKvalheim, if you're open to it... you've already suffered enough), would you be able to test this?

dithpri:
From the quick testing I've done (the repro described in 249), this works.
I think this is also much better than the solution I had suggested -- and obviously forgot about, for which I apologize -- since it avoids exporting any variables at all.

From a glance at the code, this could break compatibility with fish, as `mcfly.fish` wasn't updated to not use `MCFLY_HISTFILE`. I have however no experience with fish and just running it with the default mcfly setup has produced no errors.

cantino:
Thanks @dithpri!

I'm honestly not 100% sure if fish needs it anymore. The only code path that seems like it'd hit it for fish would be calls to `shell_history::history_file_path`, and then only if HISTFILE isn't set.

PR #309, opened 2022-12-10
Revert #216 and use which instead

cantino:
Based on https://github.com/cantino/mcfly/issues/292 and @efreak's analysis in https://github.com/cantino/mcfly/issues/292#issuecomment-1328569552, we're reverting #216 and just using `which` again instead of `command -v`.


cantino:
Should fix #292.

PR #321, opened 2022-12-30
Add forward-compatibility check for database schema

bnprks:
This adds an extra safety check in case mcfly encounters a schema newer than it knows how to deal with. 

The main motivation is to provide safety for users of my [`mcfly-fzf`](https://github.com/bnprks/mcfly-fzf) companion tool. This way users will see an obvious error if the DB schema used by their `mcfly` binary somehow is off from the schema used by their `mcfly-fzf` binary.

### DIFF
--- a/src/history/schema.rs
+++ b/src/history/schema.rs
@@ -24,7 +24,7 @@ pub fn migrate(connection: &Connection) {
 
     if current_version > CURRENT_SCHEMA_VERSION {
         panic!(
-            "McFly error: Database schema version ({}) is newer than the max version supported by this binary ({})",
+            "McFly error: Database schema version ({}) is newer than the max version supported by this binary ({}). You should update mcfly.",
             current_version,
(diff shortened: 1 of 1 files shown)


src/history/schema.rs
```diff
@@ -22,6 +22,14 @@ pub fn migrate(connection: &Connection) {
         .unwrap_or_else(|err| panic!("McFly error: Query to work ({})", err))
         .unwrap_or(0);
 
+    if current_version > CURRENT_SCHEMA_VERSION {
+        panic!(
+            "McFly error: Database schema version ({}) is newer than the max version supported by this binary ({})",
```
cantino, line 27: Maybe add "You should update mcfly." ?

bnprks: Good suggestion, just updated the message

cantino:
Thanks @bnprks

cantino:
Released in v0.8.0.

Issue #267, opened 2022-06-09
Support for per-directory-history?


ToontjeM:
Would it be possible have support for https://github.com/jimhester/per-directory-history in McFly?

ToontjeM:
Ah, i guess not, since "McFly stores its SQLite database in the standard location for the OS."

jankatins:
This should be doable: There is https://github.com/ddworken/hishtory which stores per directory in a sqllite db:

> hishtory is a better shell history. It stores your shell history in context (**what directory you ran the command in**, whether it succeeded or failed, how long it took, etc). This is all stored locally and end-to-end encrypted for syncing to to all your other computers. All of this is easily queryable via the hishtory CLI. This means from your laptop, you can easily find that complex bash pipeline you wrote on your server, and see the context in which you ran it.
> ...
> For true power users, you can even query in SQLite via `sqlite3 -cmd 'PRAGMA journal_mode = WAL' ~/.hishtory/.hishtory.db`.

mbrav:
As a long time user of `mcfly`, I just realized how useful this feature would be. It should not be hard to implement, just add another table in the SQLite database that keeps track of the current directory, a couple code additions and voila. I might consider picking this up as a side project and see I can get a viable PR. I really enjoy using `mcfly` and install it virtually on all Linux installations I interact with.

PR #322, opened 2023-01-06
feat: per directory history

rawkode:
- Adds F3 as a switcher from Global to Current Directory
- MCFLY_RESULTS_FILTER env can be GLOBAL or CURRENT_DIRECTORY

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -10,7 +10,7 @@ use std::{fmt, fs, io};
 use crate::history::{db_extensions, schema};
 use crate::network::Network;
 use crate::path_update_helpers;
-use crate::settings::{HistoryFormat, ResultSort, Settings};
+use crate::settings::{HistoryFormat, ResultFilter, ResultSort, Settings};
 use crate::simplified_command::SimplifiedCommand;
 use itertools::Itertools;
 use rusqlite::types::ToSql;
@@ -241,6 +241,8 @@ impl History {
         num: i16,
         fuzzy: i16,
         result_sort: &ResultSort,
+        result_filter: &ResultFilter,
+        dir: &String,
     ) -> Vec<Command> {
         let mut like_query = "%".to_string();
 
@@ -257,6 +259,11 @@ impl History {
             _ => "rank",
         };
 
+        let dir_filter = match &result_filter {
+            ResultFilter::Global => "%".to_string(),
+            ResultFilter::CurrentDirectory => dir.to_string(),
+        };
+
         let query: &str = &format!(
             "{} {} {} {}",
             "SELECT id, cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
@@ -264,7 +271,7 @@ impl History {
                 selected_dir_factor, dir_factor, overlap_factor, immediate_overlap_factor,
                 selected_occurrences_factor, occurrences_factor, last_run
             FROM contextual_commands
-            WHERE cmd LIKE (:like)",
+            WHERE cmd LIKE (:like) AND dir LIKE (:dir)",
             "ORDER BY",
             order_by_column,
             "DESC LIMIT :limit"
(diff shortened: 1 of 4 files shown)


rawkode:
Fixes #267

src/history/history.rs
```diff
@@ -257,14 +259,19 @@ impl History {
...
+
         let query: &str = &format!(
             "{} {} {} {}",
             "SELECT id, cmd, cmd_tpl, session_id, when_run, exit_code, selected, dir, rank,
                 age_factor, length_factor, exit_factor, recent_failure_factor,
                 selected_dir_factor, dir_factor, overlap_factor, immediate_overlap_factor,
                 selected_occurrences_factor, occurrences_factor, last_run
             FROM contextual_commands
-            WHERE cmd LIKE (:like)",
+            WHERE cmd LIKE (:like) AND dir LIKE (:dir)",
```
cantino, line 274: When `result_filter` is `ResultFilter::Global`, can we not include the `AND dir LIKE (:dir)` SQL at all? This may be faster than `AND dir LIKE "%"`.

src/interface.rs
```diff
@@ -76,7 +77,15 @@ impl MenuMode {
             ResultSort::LastRun => menu_text.push_str("F1 - Switch Sort to Rank | "),
         }
 
-        menu_text.push_str("F2 - Delete");
+        menu_text.push_str("F2 - Delete | ");
+
+        match interface.result_filter {
+            ResultFilter::Global => menu_text.push_str("F3 - Show Directory Command History"),
```
cantino, line 83: This text may be too long and overflow the width of many people's terminal windows. Maybe `F3 - This directory` and `F3 - All directories` or something?

src/interface.rs
```diff
@@ -408,6 +420,13 @@ impl<'a> Interface<'a> {
         }
     }
 
+    fn switch_result_filter(&mut self) {
+        match self.result_filter {
```
cantino, line 424: Is this the same as

```rust
self.result_filter = match self.result_filter {
    ResultFilter::Global => ResultFilter::CurrentDirectory,
    ResultFilter::CurrentDirectory => ResultFilter::Global,
}
```

Might be more readable?

cantino (changes requested):
Thanks @rawkode, this is a cool addition.

src/history/history.rs
rawkode: I wasn't really sure of a clean way to do so without bringing in some conditionals around `named_params! { ":like": &like_query, ":limit": &num, ":dir": &dir_filter},` and that felt like it made things more difficult to understand.

Any ideas?

cantino: I guess we should benchmark it. If you see no speed difference on a large database adding or omitting `AND dir LIKE (:dir)`, then this is fine. If there is one, we may want to add the conditional.

rawkode: I actually think the query planner will remove the redundant constraint. I’ll confirm on Monday

rawkode: The query planner does not remove our `WHERE dir LIKE '%"`, which is disappointing.

Benchmarks for 10k command history:

WITH LIKE '%": Run Time: real 0.083 user 0.014999 sys 0.014527
WITHOUT LIKE: Run Time: real 0.080 user 0.016174 sys 0.014996

At 10k entries, it's barely noticeable.

Tested with:

```shell
sqlite3 ~/Library/Application\ Support/McFly/history.db

.timer ON
 SELECT * FROM commands WHERE;
 SELECT * FROM commands WHERE dir LIKE '%';
```

cantino: Out of curiosity, I wonder if `WHERE (:true_or_false OR dir LIKE '%')` would work and be faster, where `:true_or_false` is a flag when in directory mode.

cantino:
This needs a rebase.

cantino:
@rawkode Sorry that this dragged out. Want me to make that one change and merge it?

Issue #294, opened 2022-11-01
Chain multiple commands


devnoname120:
It would be great if there was an option to chain multiple commands a bit like https://github.com/barthr/redo but only with a hotkey instead.

**Suggestion of implementation**:
- Add a new hotkey <kbd>Shift ⇧</kbd> <kbd>Tab ↹</kbd> that selects entries instead of switching to edit mode.
- Pressing the <kbd>Return ↩</kbd> key would then chain the command like this:

    ```shell
    command1; command2; command3;
    ```
    or alternatively:
    ```shell
    command1 && command2 && command3
    ```

sou-chon:
I can take a look at this

PR #327, opened 2023-01-27
Add mode for chaining multiple commands

YuviTz1:
Added a new mode `ChainCommands` which can be selected using `Shift + Tab`.  
Further pressing `Shift + Tab` will chain commands.
Pressing `Tab` will shift to edit mode and `enter` key can be used to execute the formed chain.
Possible implementation for #294. 
Do tell me if something needs to be changed :)


cantino:
Cool idea @YuviTz1! It seems to work for me.

I think when you first hit shift + tab, whatever command you're on should be added to the chain right away. Seems like you have to hit it twice right now.

Also, how do you remove something from the chain?

The highlight color is very hard to read though:

image

YuviTz1:
Update: 

Going into chain mode adds selected command to the chain.

Press F3 to remove commands from the chain. However do note that previously executed chains will be treated as a single command in the new chain when removing. Example:
image


Any suggestions for Highlight color?

cantino:
Needs to be rebased.

YuviTz1:
I have rebased the changes but now if the command chain takes up more than one line in the menubar, Only the first line updates successfully when removing commands from the chain for example-
image
Here, after removing the last command from the chain, 2nd line of menubar is not refreshed.
This behaviour was not present before rebasing.

cantino:
I think the way that we cleared lines got changed with @jtschuster's upgrade. Do you want to either truncate at the screen width or make this show up on its own line below the header?

Issue #284, opened 2022-09-13
feature request: selecting places on line instead of executing


samm81:
hi @cantino ! thanks for your work on this, super neat. I know you're only maintaining part time, but I figured I'd throw this out there.

I'd love it if mcfly didn't execute the command when I selected it, but instead placed it on the command line for viewing/editing, allowing me to hit enter and execute it myself once I'd verified/changed it. or for there to be a flag that could turn on this behavior.

thanks again for your work!

navazjm:
@samm81 When you select a command, instead of pressing the enter key, press tab and it should place it on the command line and not execute it

samm81:
yay, awesome! just what I wanted :) this is nowhere in the docs as far as I can tell - perhaps this issue will help others find this functionality :)

cantino:
Glad you like it @samm81! Thanks @navazjm :)

PR #334, opened 2023-03-02
Add option to disable command running

chaserhkj:
This patch adds a new option `MCFLY_DISABLE_RUN_COMMAND`, which disables all command runs in mcfly when enabled.

This is related to #284 - I have been using fzf-history for quite a while and have built muscle memory of using enter to select a history command line and edit it. Since I switched to mcfly, the new enter key to select and execute a command may accidentally fire up unexpected command runs. This had made a lot of close calls to me already - dangerous history commands like `rm` and removal API calls.

It's really hard to adapt after years of usage. So I think it might make sense to add an option to mcfly that could globally turn off any command runs to avoid any surprises. This should help others who are migrating from fzf-hisory as well.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -69,7 +69,11 @@ impl MenuMode {
             }
         }
 
-        menu_text.push_str("⏎ - Run | TAB - Edit | ");
+        if interface.settings.disable_run_command {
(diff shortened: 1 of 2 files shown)


praveenperera (approved):

cantino:
Released in v0.8.0.

PR #336, opened 2023-03-06
Write over instead of clearing to avoid flicker

jtschuster:
When clearing the entire screen every key press, the header sometimes flickers as the terminal processes the clear before the next write. Instead, write over the last frame, and only clear when needed.


cantino:
Seems to work, thanks @jtschuster

PR #340, opened 2023-03-16
Added note for setting the prompt

LukasPietzschmann:
Not sure if that's a bug, but I noticed that only one-character-prompts are allowed, so I added a little note to the README.


cantino:
Thanks @LukasPietzschmann!

Issue #339, opened 2023-03-08
 BUG : Sourcing zshrc file makes mcfly ineffective


deekshithanand:
Hi there,

Issue : Mcfly works when the  new terminal is opened first time. But if for some reason, if you source your rc file (zshrc in my case) , it makes mcfly stop working. I have provided a detailed evidence below. I can confirm that this issue is reproducible on macOS iterm2 as well. 

Snippet used in my zshrc file:
```

# MCFLY configs
export MCFLY_RESULTS=20
export MCFLY_INTERFACE_VIEW=BOTTOM
export MCFLY_PROMPT="❯"
eval "$(mcfly init zsh)" 
```

Bug can be replicated as shown below :
![mcfly_bug](https://user-images.githubusercontent.com/24704695/223777198-0dfa2a26-e38e-4f95-941f-9d61cc0bc1d5.gif)

cantino:
It's odd that sourcing it again causes a problem. I think the lines at the top of mcfly.zsh may be the issue.

```zsh
if [[ -o interactive ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
  __MCFLY_LOADED="loaded"
```

I wonder if they need to be

```
if [[ -o interactive ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
  export __MCFLY_LOADED="loaded"
```

cantino:
Actually, it seems the solution is in the other direction, removing the entire `$__MCFLY_LOADED ` concept fixes it for me:

```
# Ensure stdin is a tty
if [[ -o interactive ]]; then
```

farheinheigt:
I fix the same problem by sourcing mcfly after omz in my .zshrc

PR #341, opened 2023-03-18
Allow sourcing more than once in zsh

cantino:
Hopefully fixes https://github.com/cantino/mcfly/issues/339.

@ahatzz11 and @deekshithanand, do you mind testing this?

### DIFF
--- a/mcfly.zsh
+++ b/mcfly.zsh
@@ -1,9 +1,7 @@
 #!/bin/zsh
 
 # Ensure stdin is a tty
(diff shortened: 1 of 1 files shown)


deekshithanand:
Hi @cantino , kudos for the fix , I tested from this branch and it works . I tried to source multiple times and it works as expected.
mcfly_works

cantino:
Thanks for testing it @deekshithanand!

cantino:

PR #156, opened 2021-06-03
replace termion with crossterm and make colors configurable

dmfay:
This started because I wanted to customize colors, and turned into something of a rabbit hole. To summarize:

* working with colors in termion is [pretty painful](https://github.com/Canop/broot/blob/8258d419780c08b23cd2bdcc24ff6fbbcb83273b/src/skin_conf.rs#L32-L85)
* the [relevant feature request](https://gitlab.redox-os.org/redox-os/termion/-/issues/166) in termion is two years old, and termion itself isn't showing much movement
* meanwhile, [crossterm](https://github.com/crossterm-rs/crossterm) seems like a natural successor to termion, actively improving on a very similar API

So: the code looks for ~/.mcfly/mcfly.toml and reads colors and other settings from it, defaulting to the original styling and defaults if the file doesn't exist. Environment variables are still supported as well. A self-describing sample config file is included.

This also removes one major obstacle to Windows support, since termion is *nix-only and cross-platform compatibility is a core feature for crossterm. There may be (probably are) more, but I don't have a Windows machine to try it out.

I _did_ remove `MCFLY_LIGHT` entirely, so this conflicts with #151 (sorry!). If we want multiple colorschemes, [figment](https://github.com/sergiobenitez/figment)'s switchable profiles may be useful.

PR #343, opened 2023-03-28
Added basic customisation for menubar colors

V1RE:
#156 looks promising, but hasn't seen much action in the last 2 years. So I
made a quick and simple way to customize the foreground and background colors of the menubar using env vars.

Haven't used rust too much, so lmk if I can improve on anything

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,34 @@ set -gx MCFLY_PROMPT "❯"
 ```
 Note that only single-character-prompts are allowed. setting `MCFLY_PROMPT` to `"<str>"` will reset it to the default prompt.
 
+### Custom Menubar Colors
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @V1RE! The only problem is if we merge this, we need to support these options even in an eventually, more configurable world. Perhaps thats fine though.

cantino:
Do you want to go through and wire up other color options too (text, background, highlight)?

V1RE:
Sure, what env vars should I use?

V1RE:
@cantino What to do with the lightmode setting? this would kind of make that obsolete, right?

cantino:
Can we integrate this with the new configuration system https://github.com/cantino/mcfly/pull/407?

PR #346, opened 2023-04-10
feature: per directory history

rawkode:
Rebased #322 and incorporated feedback

cc @cantino

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -42,6 +42,12 @@ pub enum ResultSort {
     LastRun,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ResultFilter {
+    Global,
+    CurrentDirectory,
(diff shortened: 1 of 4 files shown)


src/settings.rs
```diff
@@ -155,6 +163,15 @@ impl Settings {
             _ => ResultSort::Rank,
         };
 
+        settings.result_filter = match env::var("MCFLY_RESULTS_FILTER") {
+            Ok(val) => match val.as_str() {
+                "GLOBAL" => ResultFilter::Global,
```
cantino, line 168: Probably good to make this case-insensitive.

cantino:
Awesome, thank you! I'll merge this in about a week when I'm back from some travels.

cantino:
Thanks!

cantino:

PR #352, opened 2023-04-29
Remove unnecessary/unused regex features

jirutka:
To reduce binary size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ directories-next = "2.0"
 itertools = "0.10"
 libc = "0.2"
 rand = "0.8"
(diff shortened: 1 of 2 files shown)


cantino:
Does this have any other impact on our regex that I should be aware of besides removing `\d`? How significant is the size reduction?

jirutka:
> Does this have any other impact on our regex that I should be aware of besides removing \d?

No, not on the regexes you currently use.

> How significant is the size reduction?

About 400–500 kiB for the release build.

cantino:
Thanks!

cantino:

PR #353, opened 2023-04-29
Add feature flag sqlite-bundled

jirutka:
To allow linking mcfly with system-provided sqlite.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ unicode-segmentation = "1.9"
 
 [dependencies.rusqlite]
 version = "0.28"
-features = ["bundled", "functions", "unlock_notify"]
+features = ["functions", "unlock_notify"]
 
 [dependencies.crossterm]
 version = "0.26"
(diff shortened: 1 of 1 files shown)


cantino:
Thanks @jirutka, can you explain how to use this? I haven't used Rust feature flags.

Cargo.toml
```diff
@@ -41,3 +41,6 @@ features = ["use-dev-tty"]
 version = "4"
 features = ["derive"]
 
+[features]
+default = ["sqlite-bundled"]
+sqlite-bundled = ["rusqlite/bundled"]
```
jirutka, line 46: `sqlite-bundled` feature enables the `bundled` feature in `rusqlite`.

Cargo.toml
```diff
@@ -41,3 +41,6 @@ features = ["use-dev-tty"]
 version = "4"
 features = ["derive"]
 
+[features]
+default = ["sqlite-bundled"]
```
jirutka, line 45: This says that the `sqlite-bundled` feature is enabled by default, so it will behave the same way as before. I set it mainly to maintain backward compatibility, i.e. to not break existing build scripts that assume bundled sqlite. However, the default features can be disabled via `--no-default-features` build option (`cargo build --no-default-features`).

cantino:
Thanks @jirutka!

cantino:

PR #356, opened 2023-05-07
Revert "feature: per directory history"

cantino:
Reverts cantino/mcfly#346


cantino:
Reverted because it didn't work right.

cantino:
Now fixed and merged.

PR #357, opened 2023-05-07
zsh: always set MCFLY_HISTORY in main script

### LINKED ISSUE #355: Error on startup with Zellij layouts
I've recently started using Zellij with layouts. I've layouts where I would start 3 or 4 panes, each of which will just start with my shell. But from time to time, I get the following error in one of the shells:

```
mcfly_prompt_command:10: can't rename /var/folders/_s/qgmnltjj0xs4djk_s3p8mpww0000gn/T//mcfly.EYAA0qSk.new to $HISTFILE
```
It looks like while each shell has a unique `MCFLY_SESSION_ID`, they all share the same `MCFLY_HISTORY`, so my guess is that it is causing a race condition. Maybe incorporating MCFLY_SESSION_ID inside MCFLY_HISTORY can solve this issue?

Here is some more information about Zellij's startup process: https://github.com/zellij-org/zellij/issues/1061#issuecomment-1040597253.

Versions (all latest):
macOS: 13.3.1
Alacritty: v0.12.0
Zellij: v0.37.0
zsh: v5.9 (nixpkgs)
mcfly: v0.8
mktemp: one that ships with macOS (this is used to create MCFLY_HISTORY)

utkarshgupta137:
When using McFly with terminal multiplexers such as Zellij, McFly will reuse `MCFLY_HISTORY` from the parent shell causing an issue if multiple panes are launched together using layouts.
Always set `MCFLY_HISTORY` in the main shell init script to ensure that `MCFLY_HISTORY` is unique in the child shells.

Fixes #355.

### DIFF
--- a/mcfly.zsh
+++ b/mcfly.zsh
@@ -24,9 +24,7 @@ if [[ -o interactive ]]; then
   setopt interactive_comments   # allow comments in interactive shells (like Bash does)
 
   # McFly's temporary, per-session history file.
-  if [[ ! -f "${MCFLY_HISTORY}" ]]; then
-    export MCFLY_HISTORY=$(command mktemp ${TMPDIR:-/tmp}/mcfly.XXXXXXXX)
-  fi
+  export MCFLY_HISTORY=$(command mktemp ${TMPDIR:-/tmp}/mcfly.XXXXXXXX)
 
   # Check if we need to use extended history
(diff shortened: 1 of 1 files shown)


cantino:
Thanks! Hope you tested this, I'm rolling it out.

cantino:

PR #358, opened 2023-05-07
Fish private mode support

### LINKED ISSUE #354: Fish’s private mode
The question is simple: should McFly automatically respect Fish’s private mode? Because, at this point, it doesn’t, so anything you type in private mode goes to MyFly’s database, while Fish’s history stays clear of it.

The fix is simple since there’s an environment variable for that. Yet an emphasis here is on automatically. Should a user consciously decide to forbid and manually implement it in the config, or should it be enabled by default?

I’ll gladly implement the default option if the community agrees to it.

Aeron:
This PR adds Fish private mode support by updating the Fish init script. It simply forbids saving command history while in the mode.

Thus it resolves #354.

Yet it does not forbid accessing the McFly history, and it is contrary to Fish’s behavior. But it is out of #354 scope, and I do not think it is a problem since hitting <kbd>CTRL+R</kbd> is a conscious action with the expectable outcome.

### DIFF
--- a/mcfly.fish
+++ b/mcfly.fish
@@ -31,7 +31,9 @@ if test "$__MCFLY_LOADED" != "loaded"
   end
 
   function __mcfly_add_command -d 'Add run commands to McFly database' -e fish_postexec
-    # First, retain return code of last command before we lose it
+    # Check for the private mode
+    test -n "$fish_private_mode"; and return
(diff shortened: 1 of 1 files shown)


cantino:
Simple solution, thanks @Aeron!

cantino:

PR #366, opened 2023-06-04
Create draft releases in pipeline

jankatins:
Github sends out an email on creating/publishing a release (Watch -> Custom -> releases) and that email contains whatever is at that point in the release description. As the pipeline currently creates a release without a description the email is mostly empty. As the changelog is added manually later, lets only create a draft release which can be "published" after adding the changelog. The release email then contains a proper changelog entry.


cantino:
Good idea, thanks!

Issue #372, opened 2023-07-19
feature request - CLI option to dump the sqlite3 db 


Houlistonm:
Use case: Go back after a long shell session and extract the commands run to convert to an automated process:

I currently make a copy of my command db, and run commands to find a block of commands that I've run since "yesterday" or some other time range and filter using sql to output to a file.

Feature request:
Add a `dump` option to `mcfly` to print all `cmd` from the `commands` table in chronological ascending order.

Addition arguments: 
- `--before=<TIME>`
- `--since=<TIME>`
- `--rc=0`
- `--sort=[ASC|DESC]`
- `--regex=[what ever regex family is supported]`

cantino:
Cool idea. I'd be open to a PR that adds a command for this.

TD-Sky:
- What is the format of the time? RFC3339?
- Should we accept shorter format like 2023, 2023-08 etc?
- Would it be better to use `--time=<TIME>` where `<TIME>` could be `..`, `..=` or `<single-value>`?

TD-Sky:
I've got the time parser ready: [chrono-systemd-time-ng](https://github.com/TD-Sky/chrono-systemd-time-ng). I will publish the crate after implementing this request in mcfly with it.

---

@Houlistonm what is argument `--rc`?

cantino:
#374 has been merged and will be released shortly!

PR #374, opened 2023-09-04
feat(dump history): You can specify the time range, regex and format to dump

TD-Sky:
Implementation of https://github.com/cantino/mcfly/issues/372.

Now you can dump the history. For example,

```bash
$ mcfly dump > history.json
```

will dump the whole history into json file.

Here are the options:

```
      --since <SINCE>    Select all commands ran since the point
      --before <BEFORE>  Select all commands ran before the point
  -s, --sort <ORDER>     Sort order [case ignored] [default: ASC] [possible values: ASC, DESC]
  -r, --regex <REGEX>    Require commands to match the pattern
  -f, --format <FORMAT>  The format to dump in [default: json] [possible values: json, csv]
  -h, --help             Print help
```

The timestamp (since/before) format is subject to [chrono-systemd-time](https://docs.rs/chrono-systemd-time-ng/latest/chrono_systemd_time/) which is a non-strict implementation of [systemd.time](https://www.freedesktop.org/software/systemd/man/systemd.time.html).

---

I think the PR isn't in the final stage now so suggestions are welcome.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,10 +146,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "chrono-systemd-time-ng"
-version = "0.3.1"
+name = "chrono-systemd-time"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cd5f2d9ff06c270311aeb9237e2cda17034ce8219f520069dda169d97a3a55c"
+checksum = "429fae93a2fb86c5028aa38bd4a7c0c47b51eb1e30ce059c96dea25bd570dd6c"
 dependencies = [
  "chrono",
  "once_cell",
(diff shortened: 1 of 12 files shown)


cantino:
Awesome, I'll test this out shortly. Thanks @TD-Sky

cantino:
@TD-Sky would you be able to also update the `mcfly help dump` page and README with example usage and especially a timestamp example or format reference?

TD-Sky:
Let me have a try.

TD-Sky:
You could review README.md now.

Sorry for my delay, I'm busy these days.

Houlistonm:
Any plans to merge?

TD-Sky:
> Any plans to merge?

`chrono` depreciated some APIs we are using. I may need to refactor `chrono-systemd-time` to adapt the correct APIs.

TD-Sky:
Fixed.

cantino (approved):

cantino:
Released https://github.com/cantino/mcfly/releases/tag/v0.8.3

PR #379, opened 2023-11-23
Drop custom homebrew tap

tisba:
See https://github.com/cantino/mcfly/issues/122

As discussed in https://github.com/cantino/mcfly/issues/122 the custom homebrew tap can probably just be dropped. This is just a suggestion, @cantino :)

I'm not quite sure how Homebrew does updates to bottled Formula these days. `mcfly`'s Formula lives at https://github.com/Homebrew/homebrew-core/blob/master/Formula/m/mcfly.rb. There is `brew bump-formula-pr` which should probably do most of the work, I haven't used it in a while though.

I -think- all you have to do is (with the version replaced obviously):

```
brew bump-formula-pr --url https://github.com/cantino/mcfly/archive/refs/tags/v0.8.1.tar.gz mcfly
```

PR #380, opened 2023-12-03
Revert "Drop custom homebrew tap"

cantino:
Reverts cantino/mcfly#379

@praveenperera I think we need to consider how existing users who've used the tap will be informed of this change. I think we need to leave the tap files in place with a deprecation message or something.


tisba:
Sorry, didn't saw this earlier, @cantino.

Homebrew has several mechanisms to deprecate Formulae, see https://docs.brew.sh/Deprecating-Disabling-and-Removing-Formulae. In this case, I think [`deprecate!`](https://docs.brew.sh/Deprecating-Disabling-and-Removing-Formulae#deprecation) would be appropiate.

So that could look like:

```ruby
# Warning: <formula> has been deprecated because it fetches unversioned dependencies at runtime!
deprecate! date: "2024-01-05", because: "it is available in homebrew-core"
```

Issue #382, opened 2023-12-04
Can't search command with '-' arguments in PowerShell


YoshidaRyoko:
I have run "mcfly init powershell" and append its output to my PowerShell profile
It works well when I press `Ctrl+R` with no command or command with no argument, such as `cargo build`.

But when I type command like `ls -a` or  `Get-ChildItem -Name` it reports an error

```text
error: unexpected argument '-a' found

  tip: to pass '-a' as a value, use '-- -a'

Usage: mcfly.exe search [OPTIONS] [COMMAND]...

For more information, try '--help'.
``` 

The reason of this error is when pressing `Ctrl+R`, the input line is not properly quoted.

PR #383, opened 2023-12-04
fix(pwsh): fix command with argument searching

YoshidaRyoko:
Quote current input line with quotes in PowerShell's init script, then pass it as a single argument to `mcfly search`

This commit fixs issue #382


cantino:
Did you want this merged?

PR #384, opened 2023-12-04
Append history to dummy file for compatibility

Namorzyny:
After changing `HistorySavePath` to a dummy file, PSReadline can only access newer history. For compatibility, we should make a copy as a dummy file instead of a new file.


cantino:
Thanks!

PR #385, opened 2023-12-04
fix(pwsh): fix command with argument searching

### LINKED ISSUE #382: Can't search command with '-' arguments in PowerShell

YoshidaRyoko:
Quote current input line with quotes in PowerShell's init script, then pass it as a single argument to `mcfly search`
fix #382


cantino:
Thanks!

PR #392, opened 2023-12-25
Fix draft releases breaking CD

### LINKED ISSUE #391: Draft releases messed up
@praveenperera it looks like each package is going into a different draft release now, I don't think due to any change on our part. Have you seen this before? https://github.com/cantino/mcfly/releases

praveenperera:
Fixes #391 @cantino


cantino:
Thank you!

PR #393, opened 2023-12-30
feat: add paste from clipboard support in emacs mode

eatradish:
Allow user use clipboard paste in mcfly emacs mode.

Pasting in vim mode is not supported at the moment, as it seems to require the introduction of the `copypasta` dependency in order to do that.


cantino:
Thanks!

cantino:

PR #394, opened 2024-01-06
style: apply `cargo clippy` suggest

eatradish:


cantino:
Should be fixed now, thanks!

PR #405, opened 2024-03-28
Allow colors to be configured via config file

exokernel:
## What?

Allow the colors to be configured via `~/.config/mcfly/config.toml`. This can also serve as a starting point for other config options. It has settings for both dark-mode and light-mode colors.

Example (my current color config for dark background):

```toml
[colors]

[colors.menubar]
bg = "black"
fg = "red"

[colors.darkmode]
prompt = "cyan"
timing = "yellow"
results_selection_fg = "cyan"
results_selection_bg = "black"
results_selection_hl = "red"
```

## Why?

I love mcfly but the default colors aren't working for me. I thought it might be nice if they could be set in a config file. This adds that capability by reading a basic `colors` config in a `~/.config` file for mcfly. It's written such that any errors or missing config will silently just fall back to using the defaults, which I've kept the same.

## Caveats

Reads the config file each time mcfly is run so that does add some overhead. An alternative might be to have different hard-coded color modes that could be enabled by a command line flag. This would mean no file read but also that you'd be limited to the set of color modes that are supported. The Colors struct used in the project seems pretty limited to begin with though (i.e. there's not a huge number of options). Some colors also didn't work for me in some contexts (e.g. setting the menubar to certain colors didn't always work for me).

## Thoughts and Feedback

Please let me know if you have thoughts on better ways to do this, if you'd like to see more extensive testing, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-trait"
+version = "0.1.76"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "531b97fb4cd3dfdce92c35dedbfdc1f0b9d8091c8ca943d6dae340ef5012d514"
+dependencies = [
+ "proc-macro2 1.0.70",
+ "quote 1.0.33",
+ "syn 2.0.39",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -123,6 +134,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.21.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


cantino:
Thank you for this PR. Making a few changes in https://github.com/cantino/mcfly/pull/407

cantino:
Merged in https://github.com/cantino/mcfly/pull/407, thanks @exokernel!

cantino:

cantino:
@jtschuster @torculus do you know?

torculus:
@gityhubi @cantino PowerShell [automatically expands `~` to the user's home directory](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_special_characters?view=powershell-7.4#tilde-), i.e. `C:\Users\your_name`, so using `~/.config/mcfly/config.toml` should work perfectly fine 👍

exokernel:
When this was merged in #407 I think the directory ended up being different than ~/config/mcfly/config.toml. I originally used that bc it's similar to what I've seen used in other Rust packages but mcfly uses XDG directories which may differ on different OSes iiuc. I am not sure of the top of my head what the directory is since it has been a while since I have looked at this. I'll do some poking and get back to you.

exokernel:
@gityhubi  So the code that determines the config path is

```rust
    // Use ~/.mcfly only if it already exists, otherwise create 'mcfly' folder in XDG_DATA_DIR
    #[must_use]
    pub fn mcfly_config_path() -> PathBuf {
        let data_dir = Settings::mcfly_xdg_dir().data_dir().to_path_buf();

        Settings::mcfly_base_path(data_dir).join(PathBuf::from("config.toml"))
    }
```

The documentation for `data_dir()` says:

```
Returns the path to the project's data directory.
Platform | Value | Example
-- | -- | --
Linux | $XDG_DATA_HOME/_project_path_ or $HOME/.local/share/_project_path_ | /home/alice/.local/share/barapp
macOS | $HOME/Library/Application Support/_project_path_ | /Users/Alice/Library/Application Support/com.Foo-Corp.Bar-App
Windows | {FOLDERID_RoamingAppData}\_project_path_\data | C:\Users\Alice\AppData\Roaming\Foo Corp\Bar App\data
```

And `mcfly_base_path` uses your homedir. So my guess is that on Windows this will be somewhere in `C:\Users\username\AppData\...mcfly\data\` or something similar to that.

I haven't tested this on a real system but if you are still having trouble reply back here and I will try and do that. I want this to work for you! Let me know how it goes.

torculus:
Yeah, I can confirm that there is a `C:\Users\your_name\AppData\Local\McFly\data` directory, but putting the config there doesn't work

But creating the `C:\Users\username\.mcfly\config.toml` route DOES work! 🙂

PR #408, opened 2024-05-02
Update README with Windows/pwsh instructions, and use data_local_dir to use %LOCALAPPDATA% on Windows

jtschuster:
Adds instructions for installing for Windows, and configuring for powershell.

Use [`data_local_dir`](https://docs.rs/directories-next/latest/directories_next/struct.ProjectDirs.html#method.data_local_dir) instead of [`data_dir`](https://docs.rs/directories-next/latest/directories_next/struct.ProjectDirs.html#method.data_dir) to use `C:\Users\username\AppData\Local` instead of `C:\Users\username\AppData\Roaming` to avoid syncing large DBs across machines. The path remains the same on Linux and Mac.

Also, updates references to OS X to macOS.


cantino:
Thank you @jtschuster. Do we need to check data_dir in case the user already has a mcfly data dir there?

cantino:
Is this ready to go?

jtschuster:
> Is this ready to go?

Yes, it is!

cantino:
Thanks!

cantino:

PR #410, opened 2024-05-04
Wrong order of commands in mcfly.fish

manfredlotz:
Fixes wrong order of commands in __mcfly_add_command which prevented to retain return code of last command.


cantino:
Thanks @manfredlotz!

cantino:

PR #414, opened 2024-05-16
fix init issue for fish

4t8dd:


cantino:
Released in https://github.com/cantino/mcfly/releases/tag/v0.8.6, thanks @4t8dd!

Issue #404, opened 2024-03-26
Panic in `src/fake_typer.rs` - broken edit and run commands 


nuke-web3:
Happy user that has somehow broken the tool. Not sure how to troubleshoot without building and running a debugger locally that I hope not to have to do...

### Steps to reproduce

- empty or partial start on command in terminal
- `crtl+r` brings up the gui to search
- tab or enter on selection
- ... back to an empty terminal (erased anything prompted up front)

I reinstalled with `cargo install mcfly` and followed from https://github.com/cantino/mcfly?tab=readme-ov-file#install-manually-from-source , now using mcfly 0.8.4, updated from a previously working (until mysteriously broke) 0.8.2

NukeManDan:
Placing some println in main:

```sh
nuke@pop-desk:~/git/installs/mcfly$ cargo r -- -d search -o /tmp/mcfly.asdf git
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/mcfly -d search -o /tmp/mcfly.asdf git`
=========
Selection
Some("cd git/")
=========
Settings
Settings { mode: Search, debug: true, fuzzy: 0, session_id: "xNYnvLp02hQOno1hoN1i7YE1", mcfly_history: "/tmp/mcfly.Cm4SEg8n", output_selection: Some("/tmp/mcfly.asdf"), command: "git", dir: "/home/nuke/git/installs/mcfly", results: 10, when_run: None, exit_code: None, old_dir: None, append_to_histfile: None, refresh_training_cache: false, lightmode: false, key_scheme: Emacs, history_format: Bash, limit: None, skip_environment_check: false, init_mode: Bash, delete_without_confirm: false, interface_view: Top, result_sort: Rank, result_filter: Global, disable_menu: false, prompt: "$", disable_run_command: false, time_range: TimeRange { since: None, before: None }, sort_order: Asc, pattern: None, dump_format: Json }
nuke@pop-desk:~/git/installs/mcfly$ cat /tmp/mcfly.asdf
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /tmp/mcfly.asdf
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ mode display
   2   │ commandline cd git/
───────┴────────────────────────────────
```

So output works, but there is a **panic** in default mode:

```sh
cargo r -- -d search git
[...]

PR #416, opened 2024-05-21
Avoid TIOCSTI on Bash

jtschuster:
To avoid using TIOCSTI on bash, mcfly can utilize the readline bindings. However, `bind` only lets you bind _either_ a bash command / function _or_ a readline function, but we can bind Ctrl+R to two keybindings - one which runs mcfly and binds/unbinds the second key to the `accept-line` readline depending on if `enter` or `tab` is pressed. It's a bit of a hack, so there's an escape hatch if a user prefers to use TIOCSTI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,6 +443,10 @@ McFly stores its SQLite database in the standard location for the OS. On OS X, t
 
 If you have a very large history database and you notice that McFly launches slowly, you can set `MCFLY_HISTORY_LIMIT` to something like 10000 to limit how many records are considered when searching. In this example, McFly would search only the latest 10,000 entries.
 
+### Bash TIOCSTI
+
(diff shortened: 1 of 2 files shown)


cantino:
Fascinating approach, I'll have to play with this and get back to you, thank you @jtschuster!

cantino:
It seems to basically work in bash on Mac and Linux. On Linux I am getting the following error during script setup: `bash: [: =: unary operator expected`

jtschuster:
Thanks for the heads up, missed that in my testing. Looks like it was an issue with using single brackets in the if statements, should be fixed now.

nuke-web3:
In #404 running `cargo r -r -- -d search git` for this branch brings up the TUI to select history correctly, but with `sudo sysctl -w dev.tty.legacy_tiocsti=0` set now `tab` does not fill the command in the tty. (same behavior as before, I have not debuged if it's also from the same panic)

The same code referencing TIOCSTI exists in this branch that was panicing, so I am not sure this is a complete fix?

https://github.com/cantino/mcfly/blob/caad8168c407ee239e61ec10e4e02101a77cd0e4/src/fake_typer.rs#L10-L19

cantino:
@nuke-web3 this will only work if executed from bash using the key bindings, I think.

jamedeus:
Works on Ubuntu 24.04 server and desktop (bash 5.2.21, kernel 6.8.0-31)

cantino:
Works for me in linux and mac bash.

@jtschuster do you think there's any avenue for fixing it for manual executions of mcfly search? I suspect no.

jtschuster:
No, I couldn't find any drop-in replacement for it.

cantino:
This is a good solution, thank you @jtschuster!

PR #420, opened 2024-05-30
Feature: quick stats 📊

nicokosi:
See https://github.com/cantino/mcfly/issues/330.

I propose this basic implementation. :blush: 
I introduced a few unit tests. Integration tests may be more useful.

See the [commit message](https://github.com/cantino/mcfly/pull/420/commits/7878c2aeec027a04fe84e73560bf0dd4ff6e7d73) for an example of output.

Note that the number of "first commands, sorted by occurrences" is hard-coded to 10. Should it be configurable from the command line?

PR #423, opened 2024-07-03
Auto-fix Clippy pedantic warnings

nicokosi:
Via:

    cargo clippy --fix -- -Wclippy::pedantic

After:

    cargo clippy -- -Wclippy::pedantic 2>&1| grep "warning:" | wc -l
    285

Now:

    cargo clippy -- -Wclippy::pedantic 2>&1| grep "warning:" | wc -l
    112

PS: several fixes relates to my own code changes from #420. :innocent:

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -109,7 +109,7 @@ impl<'a> Interface<'a> {
         Interface {
             history,
             settings,
-            input: CommandInput::from(settings.command.to_owned()),
+            input: CommandInput::from(settings.command.clone()),
             selection: 0,
             matches: Vec::new(),
             debug: settings.debug,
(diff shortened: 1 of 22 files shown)


src/interface.rs
```diff
@@ -866,7 +857,7 @@ impl<'a> Interface<'a> {
 
                 KeyEvent {
                     modifiers: KeyModifiers::CONTROL,
-                    code: Char('c') | Char('g') | Char('z') | Char('r'), // TODO add ZZ as shortcut
+                    code: Char('c' | 'g' | 'z' | 'r'),
```
cantino, line 860: Did you mean to remove this comment?

cantino (commented):
I assume you tested and everything still seems to work?

src/interface.rs
nicokosi: You're right, the comment should be kept.

cantino:
Thanks!

PR #425, opened 2024-07-08
Do not cancel the entire Bash/Zsh initialization on mcfly initialization failure

akinomyoga:
The shell initialiation codes contain `return 1` at the top level. However, this causes the cancellation of the entire shell initialization because those codes are supposed to be evaluated as e.g. `eval "$(mcfly init bash)"`.  This is because the invocation of `return` in the eval'ed string takes an effect in the caller context of `eval`.

For example, with the following `~/.bashrc`,

```bash
# bashrc

eval "$(mcfly init bash)"

alias myalias='foo bar baz'
```

`myalias` would not be set up when the mcfly initialization fails because the mcfly initialization calls `return 1` at the top level, which cancels the execution of `~/.bashrc` (instead of just canceling the mcfly initialization).

This PR fixes the problem by enclosing the entire mcfly initialization in a shell function (or an anonymous function in Zsh) and allows returning from the enclosing function.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -1,8 +1,13 @@
 #!/bin/bash
 
-# Ensure stdin is a tty
-# Avoid loading this file more than once
-if [[ -t 0 ]] && [[ "$__MCFLY_LOADED" != "loaded" ]]; then
+function mcfly_initialize {
+  unset -f "${FUNCNAME[0]}"
+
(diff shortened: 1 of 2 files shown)


cantino:
Thank you @akinomyoga!

cantino:
https://github.com/cantino/mcfly/releases/tag/v0.9.1

PR #427, opened 2024-07-11
Fix misc issues in Bash

akinomyoga:
See the respective commits and their titles. I think they are relatively clear fixes, but if you need explanations, please let me know. I'll explain it.


cantino:
Thank you @akinomyoga!

akinomyoga:
Thank you! I still have a few planned PRs for style adjustments, some other improvements, etc. I'll later submit them.

PR #428, opened 2024-07-12
Initialize McFly only in interactive shell

akinomyoga:
In the current setup for Bash and Fish, only keybindings are disabled in non-interactive shells, but the other settings (including `PROMPT_COMMAND`, `MCFLY_SESSION_ID`, etc.) are still loaded. Do they need to be loaded in the non-interactive shells? On the other hand, in the Zsh setup, they are not enabled in non-interactive shells because there is a check `[[ -o interactive ]]` at the beginning of `mcfly.zsh` (though there is an extra check for the interactive shell also for the keybindings in Zsh).

In this PR, I suggest skipping all the initialization in non-interactive shells to make it consistent with Zsh, which I guess is the natural behavior. I also remove the redundant check for the interactive shells in the later part of the Zsh integration.

If these settings (other than keybindings) are intentionally enabled in non-interactive shells of Bash and Fish, please let me know. I can adjust or close the PR. Thank you.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -6,6 +6,9 @@ function mcfly_initialize {
   # Ensure stdin is a tty
   [[ -t 0 ]] || return 0
 
+  # Ensure an interactive shell
+  [[ $- =~ .*i.* ]] || return 0
+
   # Avoid loading this file more than once
   [[ "${__MCFLY_LOADED-}" != "loaded" ]] || return 0
   __MCFLY_LOADED="loaded"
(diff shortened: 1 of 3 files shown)


cantino:
Thanks @akinomyoga!

akinomyoga:
Thanks! I'll submit other PRs (I waited for this PR to avoid unnecessary conflicts caused by the merging order).

In case you missed it, I replied to your email at around 21:25 on Thu, 11 Jul 2024 (PDT).

PR #429, opened 2024-07-17
Adjust styles in the Bash integration

akinomyoga:
Different styles and conventions are mixed in the Bash integration script, so I'd like to normalize them. See each commit message for each fix.


cantino:
Thanks!

akinomyoga:
Thank you!

PR #430, opened 2024-07-17
Improve misc implementation in Bash

akinomyoga:
This conflicts with #429, but after either one is merged, I'll rebase the other to resolve the conflicts.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -75,7 +75,10 @@ function mcfly_initialize {
       # conflicts with other frameworks.
       if [[ " ${PROMPT_COMMAND[*]-} " != *" $command "* ]]; then
         PROMPT_COMMAND[0]=${PROMPT_COMMAND[0]:-}
-        PROMPT_COMMAND+=("$command")
+        # Note: We here use eval to avoid syntax error in Bash < 3.1.  We drop
+        # the support for Bash < 3.0, but this is still needed to avoid parse
+        # error before the Bash version check is performed.
+        eval 'PROMPT_COMMAND+=("$command")'
       fi
     elif [[ -z ${PROMPT_COMMAND-} ]]; then
       PROMPT_COMMAND="$command"
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -128,7 +128,7 @@ function mcfly_initialize {
     #   2. Type "mcfly search" and then run the command. McFly will pull the last line from the $MCFLY_HISTORY file,
     #      which should be the commented-out search from step #1. It will then remove that line from the history file and
     #      render the search UI pre-filled with it.
-    if set -o | grep "vi " | grep -q on; then
+    if [[ -o vi ]]; then
       bind '"\C-r": "\e0i#mcfly: \e\C-m mcfly search\C-m"'
     else
       bind '"\C-r": "\C-amcfly: \e# mcfly search\C-m"'
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -27,7 +27,12 @@ function mcfly_initialize {
   export MCFLY_SESSION_ID
 
   # Find the binary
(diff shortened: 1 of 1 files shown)


cantino:
Are there any issues with bash 3 vs 4 vs 5 with these changes? I know we've had issues in the past on older systems (mostly Macs I think).

akinomyoga:
Thanks for your review. These should work in Bash 3.1+. To use the array `PROMPT_COMMAND` (which is only supported in Bash 5.1), this PR checks the Bash version on [this line](https://github.com/cantino/mcfly/pull/430/files#diff-58b7b9392506557bc74d74c8ade760aa9354f14c084e48005552f09e30bfaedfR67). The other features are all available in Bash 3/4/5. 

Bash 3.0 would fail to parse `PROMPT_COMMAND+=(...)`. If you would like to make it work also in Bash 3.0, I can adjust it. For reference, Bash bundled with macOS is Bash 3.2 and doesn't have this issue. The unmaintained MSYS1 under the MinGW project had shipped the old version of Bash 3.1 until recently, but I'm not aware of any systems that still ship Bash 3.0.

akinomyoga:
I'd suggest including a check for the Bash version at the beginning of `mcfly.bash`. What do you think? If you agree. I can add a commit in this PR.

cantino:
Yes, thank you. Maybe we just warn if it's older than 3.1? If that's on Mac and Linux by default than I don't think we need to support less than 3.1.

akinomyoga:
rebased

akinomyoga:
I finally decided to include the support for Bash 3.0. To perform the Bash version check properly, we anyway need to allow lower versions of Bash to *parse* the script. Then, Bash 3.0 actually works under such a modification.

**Details**--Because of the nature of the script that is supposed to be evaluated by `eval "$(mcfly init bash)"`, we cannot perform the *early return* from the script based on the Bash version. So the entire script needs to be correctly *parsed* even when the script is evaluated in Bash lower than 3.1 and would be disabled by the Bash version check. Meanwhile, the only thing we need to modify for Bash 3.0 is to allow Bash 3.0 to parse the part only enabled in Bash 5.1, which is the same as the modification for the Bash version check.

mcfly.bash
```diff
@@ -27,8 +34,8 @@ function mcfly_initialize {
   export MCFLY_SESSION_ID
 
   # Find the binary
-  MCFLY_PATH=${MCFLY_PATH:-$(command which mcfly)}
-  if [[ -z $MCFLY_PATH ]]; then
+  MCFLY_PATH=${MCFLY_PATH:-$(builtin type -P mcfly)}
```
cantino, line 37: We ran into an issue in the past with which, see https://github.com/cantino/mcfly/issues/292#issuecomment-1328569552 and https://github.com/cantino/mcfly/pull/309. Does builtin fix this?

akinomyoga: I do not seem to be able to reproduce the original problem of `./bin/mcfly` in a normal setup.

However, I could reproduce the problem by adding `./bin` in `PATH`. In this case, while `command -v` returns `./bin/mcfly`, `which` (GNU which) seems to return the absolute path `/bin/mcfly`. I also checked the behavior of `builtin type -P mcfly`, which produces `./bin`. To me, this appears to be a user error of including a relative path in `PATH`.

Anyway I can add a workaround in a different way without using non-POSIX `which`.

akinomyoga: I fixed up the corresponding commit: 0c0e0b52. I also described the background in the commit message.

cantino: Thanks, and sorry for the delay!

cantino:
Thanks @akinomyoga!

akinomyoga:
Thanks!

aminya:
This has introduced a bug. See #435

PR #434, opened 2024-08-12
Bind to both emacs and vi-insert keymaps in Bash

akinomyoga:
Thank you for reviewing all the PRs that I submitted so far!

This is the final PR. Unlike the previous ones, this is a suggestion for a change.

**Problem**: Bash has several keymaps depending on the current editing mode (i.e., the emacs editing mode or the vi editing mode). The default keymap for the vi editing mode is separate from the one for the emacs editing mode. Currently (in the `master` branch), the Bash keybindings are only set up *in the current keymap*. If the user changes the editing mode after `eval "$(mcfly init bash)"`, the keybindings become unavailable because the default keymap is switched.

For example, with the following bashrc, the keybindings are inactive:

```bash
# bashrc
eval "$(mcfly init bash)"
set -o vi
```

**Solution**: This PR registers the keybindings to both keymaps by explicitly specifying a keymap to register the keybindings.

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -141,11 +141,14 @@ function mcfly_initialize {
   if ((BASH_VERSINFO[0] >= 4)); then
     # shellcheck disable=SC2016
     if [[ ${MCFLY_BASH_USE_TIOCSTI-} == 1 ]]; then
-      bind -x '"\C-r": "mcfly_search_with_tiocsti"'
+      bind -m emacs     -x '"\C-r": "mcfly_search_with_tiocsti"'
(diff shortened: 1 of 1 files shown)


cantino:
Thanks!

akinomyoga:
Thank you for taking the time to review and process so many pull requests from me! I appreciate your dedication to maintaining the project! Thank you.

akinomyoga:
Ah, I'm sorry for the extra `fi`.

cantino:
No problem at all, and thank you for all the time you put into these!

cantino:
https://github.com/cantino/mcfly/releases/tag/v0.9.2

PR #438, opened 2024-08-15
Only register key press in vim mode.

jtschuster:
This is in the emacs function but not vim. We previously register both key down and key up, causing two letters to be typed at a time.


cantino:
Thanks @jtschuster!

cantino:

PR #445, opened 2024-11-16
Remove unnecessary config crate features

jirutka:
Keep just toml format.

This significantly reduces the number of transitive dependencies and reduces binary size by ~300 kiB.


cantino:
Nice, thanks!

cantino:

Issue #443, opened 2024-10-08
could not init powershell


1925381584:
Hi, it could not init powershell in my machine.

commit id : 268ee06
OS: Windows 10.0.19045
Memory: 15.7 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics || Intel Corporation || Intel driver

![image](https://github.com/user-attachments/assets/d71d5544-6d56-4590-8dbb-b7e9fa1309ab)

EmileDirigera:
That is because you are running PowerShell 5.1. The script if build for version 7 or later which added support for new operators.

In version 7.4 of PowerShell, `Set-PSReadLineOption` from PSReadLine added the parameter `PredictionSource` which does not exist in 5.1 and McFly does not seem to work properly in 5.1 from my very short testing on version 9.1 (the version that is available via winget at the time of writing this).

My advise, and what I will do, is to install PowerShell Core LTS (7.4) and run McFly there. 

If you still want to use McFly, here is a modifed version of the PowerShell script that initiates McFly. 

```posh
#!/usr/bin/env pwsh

$null = New-Module mcfly {
    # We need PSReadLine for a number of capabilities
    if ($null -eq (Get-Module -Name PSReadLine)) {
        Write-Host "Installing PSReadLine as McFly dependency"
        Install-Module PSReadLine
    }

    # Get history file and make a dummy file for psreadline (hopefully after it has loaded the real history file to its in memory history)
    $env:HISTFILE = if($null -eq $env:HISTFILE -or "" -eq $env:HISTFILE) {(Get-PSReadLineOption).HistorySavePath} else {$env:HISTFILE}
    $psreadline_dummy = New-TemporaryFile
    # Append history to dummy file for compatibility
    Get-Content -Path $Env:HISTFILE | Out-File -FilePath $psreadline_dummy -Force
    Set-PSReadLineOption -HistorySavePath $psreadline_dummy.FullName


    $fileExists = Test-Path -path $env:HISTFILE
    if (-not $fileExists) {
        Write-Host "McFly: ${env:HISTFILE} does not exist or is not readable. Please fix this or set HISTFILE to something else before using McFly.";
        return 1;
    }

    # MCFLY_SESSION_ID is used by McFly internally to keep track of the commands from a particular terminal session.
    $MCFLY_SESSION_ID = new-guid
    $env:MCFLY_SESSION_ID = $MCFLY_SESSION_ID

[...]

PR #446, opened 2024-12-21
Update README.md with powershell and bash versions

cantino:
https://github.com/cantino/mcfly/issues/443


cantino:
https://github.com/cantino/mcfly/pull/447

PR #451, opened 2025-01-25
Fix #448

### LINKED ISSUE #448: Search panics with `user-provided comparison function does not correctly implement a total order`
Time to time the search functionality (both triggered by `ctrl + r` or `mcfly search`) crashes with:
```
thread 'main' panicked at core/src/slice/sort/shared/smallsort.rs:860:5: user-provided comparison function does not correctly implement a total order
```

The panic comes from this part of the code: https://github.com/cantino/mcfly/blob/master/src/history/history.rs#L408-L449. 

This is the stacktrace:
```
 $ thread 'main' panicked at core/src/slice/sort/shared/smallsort.rs:860:5:
user-provided comparison function does not correctly implement a total order
stack backtrace:
   0:        0x10019e3f0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::habbf9c4f641febb1
   1:        0x10007ccac - core::fmt::write::ha36a8060c13608ea
   2:        0x100177194 - std::io::Write::write_fmt::h431832c8ebcc85c9
   3:        0x1001a1ba8 - std::panicking::default_hook::h4ebc6eb4ae179807
   4:        0x1001a2ac0 - std::panicking::rust_panic_with_hook::h6a84efe4dcab239c
   5:        0x1001a1e18 - std::panicking::begin_panic_handler::{{closure}}::h5eef292190467fef
   6:        0x1001a1db4 - std::sys::backtrace::__rust_end_short_backtrace::hd7e7925203f20af9
   7:        0x1001a1da8 - _rust_begin_unwind
   8:        0x100327284 - core::panicking::panic_fmt::h410d3f147658259b
   9:        0x100327714 - core::slice::sort::shared::smallsort::panic_on_ord_violation::h839471df44ebe912
  10:        0x1000c91d4 - core::slice::sort::stable::quicksort::quicksort::h0facaa6d199e1507
  11:        0x1000d0c34 - core::slice::sort::stable::drift::sort::h9266b210b2e53b2e
  12:        0x1000d0ae4 - core::slice::sort::stable::driftsort_main::hbef73e36ca0d9fe9
  13:        0x1000a1b90 - mcfly::history::history::History::find_matches::hf20bc79d3d9c53d5
  14:        0x1000c3b14 - mcfly::interface::Interface::display::h3d4bd061a91b7141
  15:        0x100003e98 - mcfly::main::h10d8a17722a95eef
  16:        0x1000f0cac - std::sys::backtrace::__rust_begin_short_backtrace::h88a42488dd053506
  17:        0x1000093b4 - _main
Process 86147 exited with status = 101 (0x00000065)
```

And I was also able to get values that cause the crash, see this Rust playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6dae2dc0d0e3d2788cd870c19c42ee7

I checked NaN values in the calculations and couldn't find any NaN value, and also verified that `PartialOrd::partial_cmp` at the end never falls to `unwrap_or` fallback. I suspect this is due to sort implementations in stdlib [detecting incorrect `Ord` implementations](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#new-sort-implementations) and panicking in that case starting with Rust 1.81.

cantino:
This should fix 448, I think, but I'm not sure that it's tuned right. I don't use FUZZY, so perhaps someone who does can tweak it? @unexge

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -431,19 +431,26 @@ impl History {
 
                     let a_start = a.match_bounds[0].0;
                     let b_start = b.match_bounds[0].0;
-                    let a_len = a.match_bounds[0].1 - a_start;
-                    let b_len = b.match_bounds[0].1 - b_start;
+                    let a_len   = a.match_bounds[0].1 - a_start;
+                    let b_len   = b.match_bounds[0].1 - b_start;
+                    let denom   = (a_start + b_start + a_len + b_len) as f64;
 
-                    let a_mod =
-                        1.0 - (a_start + a_len) as f64 / (a_start + b_start + a_len + b_len) as f64;
-                    let b_mod =
-                        1.0 - (b_start + b_len) as f64 / (a_start + b_start + a_len + b_len) as f64;
+                    // Fallback if everything is zero, to avoid 0/0
+                    if denom == 0.0 {
+                        return a.rank
+                            .partial_cmp(&b.rank)
+                            .unwrap_or(Ordering::Equal);
+                    }
 
-                    PartialOrd::partial_cmp(
-                        &(b.rank + b_mod * f64::from(fuzzy)),
-                        &(a.rank + a_mod * f64::from(fuzzy)),
-                    )
-                    .unwrap_or(Ordering::Equal)
+                    let a_mod = 1.0 - (a_start + a_len) as f64 / denom;
+                    let b_mod = 1.0 - (b_start + b_len) as f64 / denom;
+
+                    let a_score = a.rank as f64 + a_mod * fuzzy as f64;
+                    let b_score = b.rank as f64 + b_mod * fuzzy as f64;
+
+                    b_score
+                        .partial_cmp(&a_score)
+                        .unwrap_or(Ordering::Equal)
                 })
                 .collect();
         }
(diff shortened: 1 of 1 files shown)


unexge:
Thanks @cantino! I can apply this patch and test it for a couple of days to see if the results are good.

unexge:
So with the `main` branch, when I search for `get pods` I was getting `k get pods` as first result (which is what I expect), but with this change now it's the second match and first match seems less related (it has `pod` in it, but still I wouldn't expect it to be first):

main:
Screenshot 2025-01-27 at 11 42 17

this patch:
Screenshot 2025-01-27 at 11 43 30

I'm not very familiar with the codebase, I'll try to understand the sorting logic to see if I can tweak it to have similar results to the previous version.

Also, to overcome the original crash I just replaced `sorted_by` with `sorted_unstable_by` so it doesn't require total order and doesn't crash, with that version I think the current logic/sorting was preserved. Maybe that can be an alternative fix, but not sure if it messes up with some other cases.

cantino:
Hey @unexge, `sorted_unstable_by` seems like a much simpler solution. Was that giving you results you're happy with? If so, I'll just do that.

unexge:
Hey @cantino, yes, from my (limited) experience it's giving very similar results to the latest released version, which I was happy with.

cantino:
Okay cool, I'll merge https://github.com/cantino/mcfly/pull/454 instead.

PR #453, opened 2025-02-01
Changed the default number of results from 10 to 30

kentakom1213:
resolved issue #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ $env:MCFLY_FUZZY=2
 ```
 
 ### Results Count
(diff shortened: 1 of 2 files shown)


cantino:
Thx!

cantino:

eatradish:
After this commit,  `mcfly` takes a while to start up every time (db file size is around 12MiB), reset `MCFLY_HISTORY_LIMIT` to will fast, Maybe 30 is not a good default value?

cantino:
To clarify @eatradish, this PR didn't change things for you, but you do need to use `MCFLY_HISTORY_LIMIT` to limit search?

PR #454, opened 2025-02-02
Allow unstable sort

### LINKED ISSUE #448: Search panics with `user-provided comparison function does not correctly implement a total order`

cantino:
Fix #448 in a simpler way than https://github.com/cantino/mcfly/pull/451

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -408,7 +408,7 @@ impl History {
         if fuzzy > 0 && result_sort != &ResultSort::LastRun {
             names = names
                 .into_iter()
-                .sorted_by(|a, b| {
+                .sorted_unstable_by(|a, b| {
                     // Fuzzy matches impose new ordering criteria on top of the
                     // natural rank-based sorting: at the most basic level,
                     // shorter and earlier matches are more likely to be


unexge (approved):

workingjubilee:
image

> pub fn [sort_unstable_by](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.sort_unstable_by)<F>(&mut self, compare: F)
> where
>    F: [FnMut](https://doc.rust-lang.org/nightly/std/ops/trait.FnMut.html)([&T](https://doc.rust-lang.org/nightly/std/primitive.reference.html), [&T](https://doc.rust-lang.org/nightly/std/primitive.reference.html)) -> [Ordering](https://doc.rust-lang.org/nightly/std/cmp/enum.Ordering.html),
>
> Sorts the slice with a comparison function, without preserving the initial order of equal elements.

...

> # Panics
>
> May panic if compare does not implement a [total order](https://en.wikipedia.org/wiki/Total_order).

This does not actually seem to solve the problem.

cantino:
@unexge, does this fix the issue for you?

workingjubilee:
Because the panic will happen dynamically, and the sorts are different, if the comparison function is bad, then you can have the same data trigger one panic and not the other.

unexge:
Ah, I missed that `sort_unstable_by` can still panic. It solved the panic for me, I started using this patch ~3 weeks ago when I created the original issue and since then didn't experience any panic. Before this patch I was experiencing panics frequently.

cantino:
Let's merge it for now.

cantino:
https://github.com/cantino/mcfly/releases/tag/v0.9.3

PR #457, opened 2025-02-09
Implement case sensitivity for inputs containing uppercase letters

### LINKED ISSUE #36: Case sensitive match
Just a feature request: Similar to what [some](
https://github.com/Microsoft/vscode/issues/21362#issuecomment-282651162) [editor](
https://github.com/atom/autocomplete-plus/issues/42#issuecomment-131981261)'s autocomplete does, when the user types in all lowercase letters, behaves case insensitively by default. But if the input contains any uppercase letters, switches to case sensitive mode by giving each exactly matched character some extra points so those candidates rank higher on the list.

This is useful when the command starts with an environment variable, e.g., `RUST_LOG=...`.

kentakom1213:
fix #36 

Change to a mode that distinguishes between uppercase and lowercase when the input string contains uppercase letters.

---
[Screencast from 2025年02月10日 00時11分05秒.webm](https://github.com/user-attachments/assets/4448df78-9872-4a97-8844-45bf697428e9)

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -254,9 +254,7 @@ impl History {
         fuzzy: i16,
         result_sort: &ResultSort,
     ) -> Vec<Command> {
-        let has_uppercase = cmd.chars().any(|c| c.is_uppercase());
-
-        let (wildcard, match_function) = if has_uppercase {
+        let (wildcard, match_function) = if Self::is_case_sensitive(cmd) {
             ("*", "GLOB")
         } else {
             ("%", "LIKE")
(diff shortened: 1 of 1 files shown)


src/history/history.rs
```diff
@@ -254,30 +254,38 @@ impl History {
         fuzzy: i16,
         result_sort: &ResultSort,
     ) -> Vec<Command> {
-        let mut like_query = "%".to_string();
+        let (wildcard, match_function) = if Self::is_case_sensitive(cmd) {
```
cantino, line 257: The only potential issue I see is that I think some people (including me) have learned that they can type `%` during a search for a wildcard. I wonder if we should translate `%` into `*` when case sensitive?

cantino:
Cool approach, I like the concept a lot.

src/history/history.rs
kentakom1213: I didn't realize that feature existed! I'll try replacing `%` with `*` in `cmd` as well.

src/history/history.rs
```diff
@@ -254,30 +254,39 @@ impl History {
         fuzzy: i16,
         result_sort: &ResultSort,
     ) -> Vec<Command> {
-        let mut like_query = "%".to_string();
+        let (wildcard, match_function, cmd) = if Self::is_case_sensitive(cmd) {
+            // replace '%' with '*' for glob matching
+            ("*", "GLOB", cmd.replace("%", "*"))
```
cantino, line 259: The only weird thing now is that you can search for a literal `*` when using lowercase, but if you have any uppercase then you cannot. Should we escape literal `*` when uppercase?

kentakom1213: You're right, I hadn't considered that perspective. I'll make sure to escape '*'.

cantino:
Thanks @kentakom1213! Nice addition!

cantino:
Released in https://github.com/cantino/mcfly/releases/tag/v0.9.4. Thanks!

PR #460, opened 2025-03-22
Highlight only the matched characters

### LINKED ISSUE #455: Highlight Only the Matched Characters in Fuzzy Search
Currently, when fuzzy search is enabled, characters that are not matched are being highlighted. I would like to highlight only the matched characters.

Image

### LINKED ISSUE #459: Crash occurs when entering Japanese or non-ASCII characters in the search bar
When entering Japanese or other non-ASCII characters in the search bar of mcfly search, the program crashes.

## Steps to Reproduce
1. Run mcfly search in the terminal
1. Enter a Japanese word like `こんにちは` in the search bar
1. The program crashes

## Environment
- OS: Ubuntu 22.04, macOS Sonoma 14.2.1
- McFly version: 0.9.3

kentakom1213:
- fix #455 
- fix #459 

Previously, all characters between the first and last matched characters were highlighted. This has been fixed to highlight only the characters that actually matched.

Additionally, due to the change in the behavior of the `find_matches` function, issue #459 has been resolved as a result.

---
image

### DIFF
--- a/src/history/history.rs
+++ b/src/history/history.rs
@@ -412,16 +412,8 @@ impl History {
                     let a_start = *a.match_indices.first().unwrap_or(&0);
                     let b_start = *b.match_indices.first().unwrap_or(&0);
 
-                    let a_len = if fuzzy > 0 {
-                        a.match_indices.last().map(|i| i + 1).unwrap_or(0) - a_start
-                    } else {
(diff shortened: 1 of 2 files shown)


cantino:
Thanks @kentakom1213!

PR #483, opened 2026-03-20
Fix WinGet commands for mcfly installation

charlesroper:
- Updated WinGet installation and uninstallation commands to use the disambiguated package name

- Fix the Out-String command


cantino:
Thanks!

PR #490, opened 2026-05-25
fix(bash): respect string and array PROMPT_COMMAND on Bash 5.1+ (fixes #461)

### LINKED ISSUE #461: The array or string type of the PROMPT_COMMAND variable is not respected by `mcfly init bash` (and in fact erases it if it's an array)
I have multiple Bash shell extensions that mess with PROMPT_COMMAND:
bash-preexec, mcfly, direnv, and starship.

PROMPT_COMMAND is either a Bash array (ideally), or a string with ";" separators. The former is preferred, but the latter should be respected, if that's how it was initialized.

https://github.com/rcaloras/bash-preexec is a system to manage preexec functions similarly to zsh, by setting up a Bash array of precmd_functions that are run from PROMPT_COMMAND.

Anyway, both Starship (I will be reporting this bug there too) and mcfly do not respect a PROMPT_COMMAND that is an array; the logic checks for its existence only (not whether it's a string or an array) and simply overwrites it if [[ ! "$PROMPT_COMMAND" =~ "mcfly_prompt_command" ]] (which obviously is undefined on an array).

Direnv does this more correctly; here is the output of `direnv hook bash` on my system (NixOS):

```bash
_direnv_hook() {
  local previous_exit_status=$?;
  trap -- '' SIGINT;
  eval "$("/nix/store/l97wg9p8zazih2r45g5fbz9b93rrj7p4-direnv-2.34.0/bin/direnv" export bash)";
  trap - SIGINT;
  return $previous_exit_status;
};
if [[ ";${PROMPT_COMMAND[*]:-};" != *";_direnv_hook;"* ]]; then
  if [[ "$(declare -p PROMPT_COMMAND 2>&1)" == "declare -a"* ]]; then
    PROMPT_COMMAND=(_direnv_hook "${PROMPT_COMMAND[@]}")
  else
    PROMPT_COMMAND="_direnv_hook${PROMPT_COMMAND:+;$PROMPT_COMMAND}"
  fi
fi
```

As you can see, it respects PROMPT_COMMAND's that are Bash arrays. Bash 5.1 was the first to allow a PROMPT_COMMAND of type "bash array".

Please fix.

wyf027:
## Summary

- Detect whether `PROMPT_COMMAND` is a Bash array or a semicolon-separated string before registering `mcfly_prompt_command`.
- On Bash 5.1+, append to array hooks or prepend to string hooks without overwriting existing precmd chains.
- Use a unified duplicate check compatible with both representations.

Fixes #461.

Follows the same pattern as [direnv's bash hook](https://github.com/direnv/direnv/blob/master/gnu.bash): `declare -p PROMPT_COMMAND` distinguishes array vs string.

## Test plan

- [x] `cargo test`
- [ ] On Bash 5.1 with `PROMPT_COMMAND` as an array (e.g. bash-preexec): source `mcfly.bash`, confirm existing hooks remain
- [ ] On Bash 5.1 with string `PROMPT_COMMAND`: confirm McFly prepends without clobbering

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -76,16 +76,16 @@ function mcfly_initialize {
 
   function mcfly_add_prompt_command {
     local command=$1 IFS=$' \t\n'
+    if [[ ";${PROMPT_COMMAND[*]-};" == *";${command};"* ]]; then
+      return 0
+    fi
     if ((BASH_VERSINFO[0] > 5 || BASH_VERSINFO[0] == 5 && BASH_VERSINFO[1] >= 1)); then
-      # Bash 5.1 supports array PROMPT_COMMAND, where we register our prompt
-      # command to a new element PROMPT_COMMAND[i] (with i >= 1) to avoid
-      # conflicts with other frameworks.
-      if [[ " ${PROMPT_COMMAND[*]-} " != *" $command "* ]]; then
-        PROMPT_COMMAND[0]=${PROMPT_COMMAND[0]:-}
-        # Note: We here use eval to avoid syntax error in Bash < 3.1.  We drop
-        # the support for Bash < 3.0, but this is still needed to avoid parse
-        # error before the Bash version check is performed.
-        eval 'PROMPT_COMMAND+=("$command")'
+      if [[ "$(declare -p PROMPT_COMMAND 2>&1)" == declare\ -a* ]]; then
+        PROMPT_COMMAND+=("$command")
+      elif [[ -z ${PROMPT_COMMAND-} ]]; then
+        PROMPT_COMMAND="$command"
+      else
+        PROMPT_COMMAND="$command;${PROMPT_COMMAND#;}"
       fi
     elif [[ -z ${PROMPT_COMMAND-} ]]; then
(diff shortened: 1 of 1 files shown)


akinomyoga:
This PR is wrong in many aspects. Honestly, I'm not sure if it is meaningful to explain the reasons to the AI, but

- The AI doesn't seem to read existing discussion at all. As I explained in [#461](https://github.com/cantino/mcfly/pull/461), the direnv approach has an issue with breaking existing settings, as explained in [#461](https://github.com/cantino/mcfly/pull/461). Our approach is already a preferred one. Actually, this PR is essentially just a copy of the initial suggestion by the OP of [#461](https://github.com/cantino/mcfly/pull/461), and there is no originality.
- The AI doesn't seem to read even existing code that it removed. It ignores the explicit code comment explaining what the existing code cared about, and it introduces a syntax error in Bash < 3.1.
- This PR seems to assume that the commands added to `PROMPT_COMMAND` are always separated by semicolons, but this is a false assumption.

I feel this AI is not as clever as Copilot.

PR #491, opened 2026-05-25
fix(bash): keep Up-key history in sync when PROMPT_COMMAND runs history -a

wyf027:
## Summary
- Reload readline in-memory history from `MCFLY_HISTFILE` (`history -cn`) instead of the temporary `MCFLY_HISTORY` file after each prompt
- Keeps arrow-key navigation aligned with `~/.bash_history` when users also append via `PROMPT_COMMAND='history -a'`

Fixes #480

## Test plan
- [x] `cargo test`
- [ ] Reproduce #480: set `PROMPT_COMMAND='history -a'`, run commands, confirm Up shows latest entries

### DIFF
--- a/mcfly.bash
+++ b/mcfly.bash
@@ -68,24 +68,24 @@ function mcfly_initialize {
     #   for backwards compatibility and to load in new terminal sessions;
     # * find the text of the last command in $MCFLY_HISTORY and save it to the database.
     "$MCFLY_PATH" add --exit "${exit_code}" --append-to-histfile "${MCFLY_HISTFILE}"
-    # Clear the in-memory history and reload it from $MCFLY_HISTORY
-    # (to remove instances of '#mcfly: ' from the local session history).
-    history -cr "${MCFLY_HISTORY}"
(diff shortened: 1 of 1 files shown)


mcfly.bash
```diff
@@ -68,24 +68,24 @@ function mcfly_initialize {
     #   for backwards compatibility and to load in new terminal sessions;
     # * find the text of the last command in $MCFLY_HISTORY and save it to the database.
     "$MCFLY_PATH" add --exit "${exit_code}" --append-to-histfile "${MCFLY_HISTFILE}"
-    # Clear the in-memory history and reload it from $MCFLY_HISTORY
-    # (to remove instances of '#mcfly: ' from the local session history).
-    history -cr "${MCFLY_HISTORY}"
+    # Reload in-memory history from the canonical history file so arrow keys stay
+    # in sync when PROMPT_COMMAND also runs `history -a` against HISTFILE.
+    history -cn "${MCFLY_HISTFILE}"
```
akinomyoga, line 73: With this, couldn't the user access only the first previous command?

PR #495, opened 2026-07-13
feat: support various PSReadline if possible

### LINKED ISSUE #494: mcfly init powershell hangs in non-interactive sessions due to Install-Module PSReadLine
## Problem

`mcfly init powershell` generates a script (`mcfly.ps1`) that unconditionally calls `Install-Module PSReadLine` when PSReadLine is not already loaded:

https://github.com/cantino/mcfly/blob/master/mcfly.ps1#L13-L16

```powershell
if ($null -eq (Get-Module -Name PSReadLine)) {
    Write-Host "Installing PSReadLine as McFly dependency"
    Install-Module PSReadLine
}
```

Because `Install-Module` is called without `-Force` or `-Confirm:$false`, it prompts for user confirmation. In non-interactive sessions (CI pipelines, VS Code tasks, AI coding agents, headless terminals), there is no user to confirm, so **the shell hangs indefinitely** at `Installing PSReadLine as McFly dependency`.

This blocks the entire PowerShell profile from loading.

## Environment

- OS: Windows 11
- PowerShell: 7.x (pwsh)
- McFly: latest

## Suggested fix

Guard the install with an interactivity check, and add `-Force` so it doesn't prompt:

```powershell
if ($null -eq (Get-Module -Name PSReadLine)) {
    if ([Environment]::UserInteractive) {
        Write-Host "Installing PSReadLine as McFly dependency"
        Install-Module PSReadLine -Force -Scope CurrentUser
    } else {
        Write-Warning "McFly requires PSReadLine but cannot install it in a non-interactive session. Run 'Install-Module PSReadLine' manually."
        return
    }
}
```

## Current workaround

Wrap the mcfly init in the user's PowerShell profile with a try/catch that checks for a console window:

```powershell
try {
    if ($host.Name -eq 'ConsoleHost' -and [Console]::WindowHeight -gt 0) {
        Invoke-Expression -Command $(mcfly init powershell | out-string)
    }
} catch {}
```

aminya:
This adds handlers for various PSReadline versions with some simple checks and guards. It also fixes the bugs in non-iteractive sessions where Powershell doesn't load properly

Fixes #494

### DIFF
--- a/mcfly.ps1
+++ b/mcfly.ps1
@@ -1,27 +1,23 @@
 #!/usr/bin/env pwsh
 
 $null = New-Module mcfly {
-    # We want PSReadLine 2.2.0+ for full prediction support if available
-    $psReadLineMinVersion = [version]'2.2.0'
-    $hasNewPSReadLine = Get-Module PSReadLine -ListAvailable |
-        Where-Object { $_.Version -ge $psReadLineMinVersion }
-    if (-not $hasNewPSReadLine -and [Environment]::UserInteractive) {
-        try {
-            Write-Host "Installing PSReadLine $psReadLineMinVersion+ as McFly dependency"
-            Install-Module PSReadLine -Force -Scope CurrentUser -MinimumVersion $psReadLineMinVersion -ErrorAction Stop
-        }
-        catch {
-            Write-Warning "Could not install PSReadLine $psReadLineMinVersion+ ($($_.Exception.Message)). Continuing with the existing version."
-        }
-    }
-
-    # Prefer the newest available; fall back to whatever is already loaded.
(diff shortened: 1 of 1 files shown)


cantino:
I have no way to test this. @jtschuster, @dmfay, any opinions?

dmfay:
Supporting PowerShell is all to the good but I don't have access to a Windows machine to test either, I'm afraid. The change is pretty targeted but "delete the stale user-scoped copy" does make me nervous. I don't know whether that's an ephemeral yanking-out-of-scope or something else, and it feels like something we probably shouldn't do on the user's behalf. Safer to abort with a "requires PSReadLine 2.2.0" message?

aminya:
Hi, I will update the PR soon. I iterated over it but I didn't push the latest version

PR #497, opened 2026-08-25
Add MCFLY_HISTORY_IGNORE to skip recording matching commands

### LINKED ISSUE #86: How can I ignore some commands.
How can I ignore some commands such as exit and ls

ChrisJr404:
Closes #86. Adds an `MCFLY_HISTORY_IGNORE` environment variable so you can keep specific commands out of your history. It's a regular expression, and any command matching it won't be recorded, running right alongside the existing built-in ignore list.

While I was in there I pulled the ignore checks out of `should_add` into a small `is_ignored_command` function so the new pattern sits next to the existing rules and can be unit tested without a database. Added a few tests for it and a README section.

On the shells mentioned in the issue, I went with a dedicated mcfly variable instead of reusing `HISTIGNORE` / `HISTORY_IGNORE`, since as @nmay231 pointed out those also affect the shell's own up-arrow history, which isn't necessarily what you want here. A bad pattern just warns and is ignored so it can't stop commands from being recorded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ directories-next = "2.0"
 itertools = "0.14"
 rand = "0.9"
 path-absolutize = "3.1"
-regex = { version = "1", default-features = false, features = ["perf", "std"] }
+regex = { version = "1", default-features = false, features = ["perf", "std", "unicode-perl", "unicode-case"] }
 shellexpand = "3"
 unicode-segmentation = "1.11"
 
--- a/README.md
+++ b/README.md
@@ -455,6 +455,24 @@ $env:MCFLY_PROMPT=">"
 
 Note that only single-character-prompts are allowed. setting `MCFLY_PROMPT` to `"<str>"` will reset it to the default prompt.
 
+### Ignoring commands
+McFly already skips a few noisy commands (like `ls`, `cd`, and `pwd`) as well as anything you run with a leading space. To skip more, set `MCFLY_HISTORY_IGNORE` to a regular expression. Commands matching it will not be recorded. The pattern is unanchored, so add `^` if you want to match from the start of the command.
+
+bash / zsh:
+```bash
+export MCFLY_HISTORY_IGNORE="^(kubectl|helm) |password"
+```
+
+fish:
+```bash
+set -gx MCFLY_HISTORY_IGNORE "^(kubectl|helm) |password"
+```
+
+powershell:
+```powershell
+$env:MCFLY_HISTORY_IGNORE="^(kubectl|helm) |password"
+```
+
 ### Database Location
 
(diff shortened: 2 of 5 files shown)


cantino (changes requested):
Thank you @ChrisJr404! I think this makes sense.

Two things:
1. Can we call this `MCFLY_IGNORE_PATTERN` instead?
2. Right now Cargo.toml pins regex with `default-features = false, features = ["perf", "std"]` in order to [reduce file size](https://github.com/cantino/mcfly/commit/0bb9470), but this means no unicode and so patterns like `\d+`,   `\w+`, `\s`, `\b`, and especially `(?i)PASSWORD` fail. I think we should fix this like so:

```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@
 itertools = "0.14"
 rand = "0.9"
 path-absolutize = "3.1"
-regex = { version = "1", default-features = false, features = ["perf", "std"] }
+regex = { version = "1", default-features = false, features = ["perf", "std", "unicode-perl", "unicode-case"] }
 shellexpand = "3"
 unicode-segmentation = "1.11"
```

I also noticed that in bash, an ignored command is kept out of ~/.bash_history too, because mcfly.bash writes the histfile via --append-to-histfile and that call sits inside the should_add guard. zsh and fish write their own history, so there only mcfly's DB is filtered. So your claim that "those also affect the shell's own up-arrow history, which isn't necessarily what you want here" is true in bash, but not in the other shells. I don't know that we want to fix that here, though.

ChrisJr404:
Both done: renamed the variable to MCFLY_IGNORE_PATTERN throughout (code, README, error message), and added the unicode-perl and unicode-case regex features so patterns like `\d+`, `\b`, and `(?i)PASSWORD` work. Added a test covering the case-folding and Perl-class behavior. Good catch on the bash --append-to-histfile detail; I left that alone since the DB-side filtering is the intended scope here.

ChrisJr404:
Both done in the latest commit:

1. Renamed the variable to `MCFLY_IGNORE_PATTERN` throughout (settings, history, and README).
2. Added `unicode-perl` and `unicode-case` to the `regex` features in Cargo.toml, so `\\d`, `\\w`, `\\s`, `\\b`, and `(?i)` patterns work while keeping the size-reducing `default-features = false`.

Good point on the bash case: because `mcfly.bash` appends to the histfile inside the `should_add` guard, an ignored command is also kept out of `~/.bash_history`, whereas zsh and fish write their own history so only mcfly's DB is filtered. Agreed it is out of scope for this PR; I have left that behavior as-is.

cantino (approved):

cantino:
Thanks @ChrisJr404!

PR #499, opened 2026-09-11
Fix multiline history parsing in Bash and Zsh

### LINKED ISSUE #184: Doesn't start in new terminal windows Ubuntu 18.04
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic

When I open a new terminal window it doesn't start. I do have this is in my  ~/.bashrc
eval "$(mcfly init bash)"

I have to run . ~/.bashrc / . ~/.zshrc / source ~/.config/fish/config.fish
every time to get it working.

### LINKED ISSUE #267: Support for per-directory-history?

aoright:
Fixes #267
Fixes #184 (or similar issues where multiline commands were split).

This PR fixes the parsing of `.bash_history` and `.zsh_history` so that multiline commands are preserved correctly as a single `HistoryCommand`.

- **Bash**: When `HISTTIMEFORMAT` is in use, `#timestamp` lines denote new commands. This PR properly accumulates intermediate lines when timestamps are present.
- **Zsh**: Preserves extended history multi-line commands by checking for the `: <time>:<dur>;` prefix and correctly handling trailing backslashes to unescape newlines.


cantino (commented):
Thanks for digging into this.

Claude found some issues, however.

---

The core insight here is correct, and I verified it against real shells rather than from memory. But the change only fixes the read side of what is a read/write pipeline, and a few of the consequences are destructive, so I don't think it can merge as-is.

## Testing done

Checked out the branch: 32 tests pass, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean, CI green. Then established ground truth against zsh 5.x and bash 5.x, and ran adversarial probes against both this branch and `master`.

Ground truth confirms your model of the formats:
- zsh escapes an embedded newline as `\` + newline in **both** extended and non-extended formats.
- zsh writes a command ending in a literal backslash as `\` + space, so the continuation logic can't swallow the following entry in zsh-written files.
- `echo foo \`⏎`  bar` loses its literal backslash through zsh's own writer, so the expectation in `test_zsh_extended_multiline_commands` matches zsh's behavior.
- bash with `HISTTIMEFORMAT` emits `#<epoch>` before every entry; without it, multi-line commands have no delimiter at all and are genuinely unparseable.

## Blocking

**1. Bash + `HISTTIMEFORMAT` breaks the `#mcfly:` search marker.** `mcfly.bash` echoes `#mcfly: <buffer>` into `$MCFLY_HISTORY` with no preceding `#ts` line, so the accumulator glues it onto the previous command:

```
master: last_history_line = Some("#mcfly: carg")               → cleanup rewrites file correctly
PR:     last_history_line = Some("cargo build\n#mcfly: carg")  → file left unchanged
```

`trim_start_matches("#mcfly: ")` in `settings.rs` then no-ops, so Ctrl-R seeds the search box with `cargo build\n#mcfly: carg` instead of `carg`, and `delete_last_history_entry_if_search` stops stripping the marker — `#mcfly:` lines accumulate and get imported as history entries. This hits exactly the users the PR targets.

**2. `delete_lines` now corrupts the user's real `$HISTFILE`.** `history_cleaner::clean` rewrites the whole file by reserializing parsed commands. The parser unescapes `\`+newline → newline; `Display` never re-escapes it. Deleting a single entry from the UI turns

```
: 1593624160:0;echo "hello\ ⏎  world"
```

into a raw newline with no backslash, which real zsh then reads back as **two** entries. Same asymmetry in bash: the rewrite drops the `#ts` delimiters, so a preserved multi-line entry re-splits on the next read. Whatever `full_history` unescapes, `Display` has to re-escape — or the delete path needs to edit the raw file instead of reserializing.

**3. The non-timestamp branch dropped the `has_leading_timestamp` filter.** The old code filtered timestamp lines in both branches; the new `else` branch only filters empties. Any history whose first 10 lines predate `HISTTIMEFORMAT` being enabled — a common situation — imports the timestamps as commands:

```
["old_cmd_1", …, "old_cmd_10", "#1593624164", "echo hi"]
```

The `take(10)` heuristic is what makes this reachable. Suggestion: drop the heuristic, treat `#\d{10}` as a boundary unconditionally, and accumulate between boundaries (lines before the first boundary stay one-per-line). Same result on clean files, no heuristic, no regression.

**4. Downstream consumers assume single-line commands**, so the feature doesn't survive selection:
- `--output-selection` is line-oriented: `while read -r key val` in `mcfly.zsh`/`mcfly.fish`, `awk 'NR==2'` in `mcfly.bash`. A multi-line selection is silently truncated to its first line.
- The TIOCSTI path injects bytes one at a time, so an embedded newline is Enter — selecting a multi-line entry can execute a fragment, or auto-run the whole command without the user pressing Enter.
- `truncate_for_display` pushes raw chars into a row drawn at a fixed `MoveTo`, so a newline corrupts the result list.

Minimum viable fix: escape newlines at the selection boundary and unescape in the shell functions, or at least sanitize for display and avoid the TIOCSTI path for multi-line entries.

## Minor

- In the zsh loop, `if has_leading_timestamp(line) || line.is_empty() { continue; }` silently drops lines *inside* an accumulating command: `cat <<'X'` / `#1593624160` / `X` parses to `cat <<'X'\nX` — mangled but still runnable.
- The bash branch quietly stopped stripping zsh `: ts:dur;` prefixes. Probably intentional, but it isn't mentioned.
- `current_command.clone()` + `.clear()` → `std::mem::take(&mut current_command)` (4 sites).
- zsh's `\` + space convention for a trailing literal backslash leaves a stray trailing space in the parsed command (pre-existing, cosmetic).
- `Fixes #267` / `Fixes #184` look wrong — #267 is a per-directory-history feature request and #184 is a shell-init/PATH problem. Merging as written would auto-close two unrelated issues.

## What's good

The zsh escaping model is right, and that's the part most attempts get wrong. The `#ts`-as-delimiter approach for bash is the only workable one. Happy-path test coverage is solid and the unmetafy test is worth keeping regardless.

Happy to look at a revision — the direction is right, it mainly needs writer symmetry, boundary-based parsing, and a decision about what happens to a multi-line command at the selection boundary.

---
_Generated by [Claude Code](https://claude.ai/code)_

aoright:
Thank you @cantino for the in-depth review and detailed adversarial testing. I have pushed an update addressing all points raised:

### 1. Bash + `HISTTIMEFORMAT` and the `#mcfly:` Marker
- Added explicit boundary handling for `#mcfly:` lines in `full_history` for Bash. If a command was accumulating, it is flushed immediately and `#mcfly:` is emitted as its own entry.
- Added `test_bash_history_with_mcfly_marker` verifying that `last_history_line` returns `#mcfly: ...` and `delete_last_history_entry_if_search` removes it cleanly without mutating the preceding command or timestamps.

### 2. Writer Symmetry in `Display`
- In `fmt::Display for HistoryCommand`, Zsh multi-line commands now re-escape newlines as `\\\n` (both standard and extended formats), so reserializing during `delete_lines` preserves valid Zsh continuation syntax.
- For Bash, entries that had leading timestamps retain `has_timestamp: true` and their epoch `when`, formatting as `#{when}\n{command}` so rewriting preserves entry delimiters and multi-line grouping.
- Added `test_writer_symmetry_bash_and_zsh` verifying roundtrip fidelity through `delete_lines`.

### 3. Boundary-based Parsing without Heuristics
- Dropped the `take(10)` heuristic entirely.
- Lines prior to the first `#\d{10}` boundary remain single-line commands; subsequent lines accumulate between timestamp boundaries.
- Added `test_bash_mixed_timestamp_history` verifying histories with un-timestamped early entries do not import timestamps as commands.

### 4. Downstream Selection and UI Safety
- **Display**: `truncate_for_display` now sanitizes newlines (`\n`, `\r`) to spaces, keeping the fixed terminal row layout intact.
- **TIOCSTI**: Guarded against multiline commands in the TIOCSTI fallback path to prevent premature Enter execution.
- **Shell Integrations**: Updated selection parsers in `mcfly.bash`, `mcfly.zsh`, and `mcfly.fish` with line accumulators under `commandline `, allowing multi-line commands to populate the shell buffer completely.

### 5. Minor Cleanups
- Removed `has_leading_timestamp` filtering from the Zsh loop, allowing heredocs and embedded comments to parse intact (`test_zsh_multiline_empty_lines_and_heredocs`).
- Restored Zsh timestamp prefix stripping for imported histories in Bash.
- Replaced `.clone()` + `.clear()` with `std::mem::take(&mut current_command)`.
- Updated the PR description to remove the unrelated `Fixes #267` and `Fixes #184` references.

All 36 tests pass, and `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

